PR #56, opened 2014-07-18
added layer dialog and experiment dialog

yusufameri:


pkienzle:
Out of date. New GUI development has moved to webview.

PR #64, opened 2015-09-08
Fix Fresnel calculation

richardsheridan:
I have been concerned for a while that the plots produced when viewing data as "Fresnel" and "log Fresnel" deviated suspiciously far from 1. In trying to reproduce the plot for publication, I realized that the calculation was likely reversed, since the critical angle/wavevector 4*pi*rho was coming out negative. Then I noticed that the documentation of ables.refl doesn't match the usage in fresnel.test, i.e. incident medium should be the first item. That is to say, the test has been wrong the whole time.

If you check my work and decide I'm correct on this, perhaps we could also add some documentation to Fresnel.reflectivity to clear up what exactly are rho_Qm, rho_Qp.


pkienzle:
Yes, you are correct.  Sorry to sit on it so long...

PR #66, opened 2017-04-09
optimize slab building when step_interfaces=True

richardsheridan:
This patch improves the speed of any model that includes a microslabbed profile when the parameter step_interfaces is set to True during construction of the `Experiment` instance. Previously the computation was dominated by the `build_profile` function, rather than the reflectivity amplitude. I have observed speedup factors of 10-100 depending on the number of microslabs. A few things to note:

- This patch has no effect on any fits where `step_interfaces=False`. `build_profile` is simply not called except for plotting purposes.
- This patch does not change the calculation for any rough slabs. The blend function is simplified but the numerical result is the same.
- The patch adds an extra O(`len(thickness)`) calculation (with a small prefactor) to the beginning in exchange for reducing the rest of the calculation to O(`len(rough_layers)`)
- The patch does not add substantial time when `len(thickness)==len(rough_layers)`... about 1% +- 1%
- The quick, no-roughness profile one-liner is a bit inscrutable, but the intermediate commit shows a pure python version of the idea. 

This change was necessary to run my `EndTetheredPolymer` based fits for publication in a reasonable amount of time.

### DIFF
--- a/refl1d/profile.py
+++ b/refl1d/profile.py
@@ -62,6 +62,8 @@ using one energy so we only show the first column.
 import numpy
 from numpy import inf, nan, isnan
 from scipy.special import erf
+from itertools import izip
+sqrt2=numpy.sqrt(2.0)
 
 class Microslabs(object):
     """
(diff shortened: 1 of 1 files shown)


pkienzle:
Don't think we can do this anymore since blends now extend over all z, and are not limited to neighbouring layers.

richardsheridan:
Agreed. Closing!

PR #68, opened 2017-04-16
Stability, style, and performance tweaks for polymer self consistent field theory code.

richardsheridan:
This pull request contains all the best ideas I had on my private SCF codebase to improve stability. It's also somewhat more readable. The changes cause a slight numerical change in tests, greater than the np.allclose tolerance but tolerable in practice.

I'd appreciate if someone could build the docs off this and make sure the equations and references render properly, too.


refl1d/polymer.py
```diff
@@ -1089,4 +1095,4 @@ def _calc_g_zs_uniform(self, g_z, g_zs):
             segments = g_zs.shape[1]
             pg_zs = g_zs[:, 0]
             for r in range(1, segments):
-                g_zs[:, r] = pg_zs = correlate(pg_zs, LAMBDA_ARRAY, 1) * g_z
+                g_zs[:, r] = pg_zs = correlate(pg_zs, LAMBDA_ARRAY, 1) * g
```
pkienzle: I messed up when merging from master.  This should be `... * g_z`.  Other than that, these are only linting changes

pkienzle:
I reproduced the changes on master, so this branch can be removed.

PR #71, opened 2017-05-18
MAINT: don't use Angstrom character

andyfaff:
Leaving the Angstrom character in there seemed to cause build problems.


pkienzle:
Thanks for the update.

I have the sphinx macro ```|1/Ang|``` for representing inverse angstroms.  I did this  and pushed, so I'm closing the pull request.

There is also the option of setting ```# -*- coding: utf-8 -*-``` at the top of the file,  which is how I worked around this previously (though who knows whether my editor was storing it as utf-8 or some other variant such as iso-8859-10).

  - Paul

PR #73, opened 2018-09-17
Add option for simulate data

mdoucet:
There was a request to add more realistic errors to ```Experiment.simulate_data(noise)```, which calls ```Probe.simulate_data(theory, noise)```.

At the moment, ```noise``` is just ```dR/R``` in percents. For cases where a full ```R(q)``` curve is made of a number of smaller ```q```-range pieces acquired separately (as we have on TOF reflectometers), providing ```dR/R``` is insufficient because the statistics will vary as a function of ```q```.

The suggested change allows the ```noise``` parameter to be an array of the same length as the ```theory``` parameter.

### DIFF
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -280,6 +280,9 @@ class Probe(object):
 
         if noise is None:
             pass # use existing noise
+        elif isinstance(noise, (list, tuple, numpy.ndarray)) and len(noise) == len(theory):
+            self.dR = numpy.asarray(noise)
+            self.dR[self.dR==0] = 1e-11
         else:
             self.dR = 0.01*noise*self.Ro
             self.dR[self.dR==0] = 1e-11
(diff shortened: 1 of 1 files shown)


refl1d/probe.py
```diff
@@ -280,6 +280,9 @@ def simulate_data(self, theory, noise=None):
 
         if noise is None:
             pass # use existing noise
+        elif isinstance(noise, (list, tuple, numpy.ndarray)) and len(noise) == len(theory):
+            self.dR = numpy.asarray(noise)
+            self.dR[self.dR==0] = 1e-11
         else:
             self.dR = 0.01*noise*self.Ro
             self.dR[self.dR==0] = 1e-11
```
pkienzle: The array length test is not useful.  If the length is wrong, then it just falls into the next branch where it is used unconditionally.

I prefer to avoid isinstance() tests.  Just use numpy.asarray() and let numpy sort out the types.
 
How about:
```
if noise is None:
    pass
else:
    self.dR = np.asarray(noise)
    if len(self.dR.shape) == 0:  # noise is a scalar
        self.dR = 0.01*noise*self.Ro
    self.dR[self.dR == 0] = 1e-11
```
    
```

mdoucet: You don't mind that someone can pass an array of the wrong size?
I modified the code to use your version, and I added something similar to ```ProbeSet```.
Anywhere else that would be affected?

PR #75, opened 2018-11-06
Fix problem with Layer not having to_dict()

mdoucet:


pkienzle:
Didn't notice this in your first commit, but you are putting materials parameters and now magnetic parameters at the same level, where they can potentially collide with each other (for example, if material happened to have a parameter named 'thickness'), and where they will be hard to separate out again.

I'll pull it as is, but I think we need to rethink the scheme, and create a declarative representation of fits and parameter relations which is easier to save, load and manipulate.

PR #77, opened 2019-04-11
load4 select data range

e-rus:
Include an option in load4 to use a selected range of rows in the data file.

### DIFF
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -1041,7 +1041,7 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
           theta_offset=0, sample_broadening=0,
           L=None, dL=None, T=None, dT=None,
           FWHM=False, radiation=None,
-          columns=None,
+          columns=None,data_range=[None,None],
          ):
     r"""
     Load in four column data Q, R, dR, dQ.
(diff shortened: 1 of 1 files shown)


refl1d/probe.py
```diff
@@ -1041,7 +1041,7 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
           theta_offset=0, sample_broadening=0,
           L=None, dL=None, T=None, dT=None,
           FWHM=False, radiation=None,
-          columns=None,
+          columns=None,data_range=[None,None],
```
pkienzle: Lint will complain about spacing and about mutable default arguments.  Use the following instead:
```
columns=None, data_range=(None, None),
```

refl1d/probe.py
```diff
@@ -1116,7 +1119,8 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
     else:
         order = [0, 1, 2, 3]
     def _as_Qprobe(data):
-        Q, R, dR, dQ = (data[1][k] for k in order)
+        lower_lim,upper_lim=data_range
+        Q, R, dR, dQ = [data[1][k][lower_lim:upper_lim] for k in order]
```
pkienzle: Lint will complain about spacing: `lower_lim, upper_lim = data_range`.  

Can allow selection of every nth data point using the following:
```python
index = slice(*data_range)
Q, R, dR, dQ = [data[1][k][index] for k in order]
```

pkienzle:
Applied by hand.

PR #86, opened 2019-10-09
Bug fixes

TUM-E21-ThinFilms:
Mostly some bug fixes and two minor improvements

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,5 @@
 *.h text eol=lf
 *.cpp text eol=lf
 *.rst text eol=lf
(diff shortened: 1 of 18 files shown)


pkienzle:
I applied the changes for fit by number density and q interpolation by hand.  

The none-type access was already fixed in master.

I didn't include interfaceM for Magnetism because BaseMagnetism provides interface_below and interface_above to define the magnetic interface for a single layer.  I updated the docs.

Magnetic interfaces between layers (e.g., in MagnetismStack), are not yet supported by the profile builder, but I twiddled that code as well.

PR #91, opened 2020-05-27
ENH: add QProbe.oversample

andyfaff:


andyfaff:
@pkienzle, is this PR the correct way of going about oversampling? It would be nice to have it for the validation suite that checks if resolution smearing is done correctly. (It is, but oversampling is required for the test cases)

pkienzle:
Yes, this is fine. Unfortunately, I missed seeing it, and ended up reimplementing it in response to issue #95.

I'm adding your changes to .gitignore, so this PR is no longer needed.

PR #101, opened 2021-01-16
FIX: length of vectors not equal when dthetaM is not defined

acaruana2009:
Also fixed issue with magnetic profile calculation when drhoM[-1] = 0.

Hi Paul, could you review these tweaks? I have finally got round to testing the FreeInterfaceMagnetism function you added in a while ago. 

Essentially, the issues were:

1. When dthetaM is not defined it would only have the length of the input list, not the profile length
2. When drhoM[-1] = 0 it did not seem to calculate the profile. I have added in a check that we used in the test function I originally sent you.

Thanks

Andrew

### DIFF
--- a/refl1d/magnetism.py
+++ b/refl1d/magnetism.py
@@ -508,6 +508,10 @@ class FreeMagnetismInterface(BaseMagnetism):
         z *= thickness/z[-1]
 
         rhoM = hstack((0, cumsum(asarray([v.value for v in self.drhoM], 'd'))))
+        # AJC added since without the line below FreeMagnetismInterface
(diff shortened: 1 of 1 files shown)


bmaranville:
@acaruana2009 can you provide a test for this fix?  (a problem definition that breaks without the fix?)

acaruana2009:
@bmaranville Sure, I will put something together tomorrow. From memory, it breaks subtly i.e. you need to look at the model produced say in the GUI to see that it is not behaving right. Will this still be ok? 

The model I am aiming to send through has the node values linked to the FreeInterface (Structural SLD) so adjusting the dz and drho values should give the same shape as the FreeMagnetismInterface.

PR #106, opened 2021-02-26
update total thickness parameter to be a regular parameter instead of…

bmaranville:
… Function parameter

The implementation of this standard parameter (present in nearly every model) as a Function parameter makes it hard to serialize models.

Refactor to a simple sum of thickness parameters of all layers. (also is a parameter)

The thickness property of multilayers is also altered in this way, changing the existing behavior where multilayer.thickness -> float instead of mutlilayer.thickness -> Parameter

### DIFF
--- a/refl1d/model.py
+++ b/refl1d/model.py
@@ -606,7 +606,7 @@ class Repeat(Layer):
         pars = {
             'stack': self.stack.parameters(),
             'repeat': self.repeat,
-            'thickness': self._thickness,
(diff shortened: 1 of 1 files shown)


bmaranville:
This was implemented, but now we're going to make it into a new special type of Parameter: Calculated (meaning well-understood by the model, but not an Expression spelled out)

glass-ships:
Implemented - safe to close PR and delete branch?

PR #108, opened 2021-03-11
FIX: Make thetaM behave the same as rhoM in FreeMagnetismInterface

acaruana2009:
Removed conditions at lines 514 and 521. Also added same statement at line 520 as is used for rhoM. This should now behave in the same way as FreeInterface.

### DIFF
--- a/refl1d/magnetism.py
+++ b/refl1d/magnetism.py
@@ -510,15 +510,17 @@ class FreeMagnetismInterface(BaseMagnetism):
         rhoM = hstack((0, cumsum(asarray([v.value for v in self.drhoM], 'd'))))
         # AJC added since without the line below FreeMagnetismInterface
         # does not initialise propoerly - fixes strange behaviour at drho=0 on end point
-        if rhoM[-1] == 0: rhoM[-1] = 1
-        # if rhoM[-1] > 0:
+        if rhoM[-1] == 0:
+            rhoM[-1] = 1
+
         rhoM *= 1/rhoM[-1]
         PrhoM = clip(monospline(z, rhoM, Pz), 0, 1)
 
         if self.dthetaM:
             thetaM = hstack((0, cumsum(asarray([v.value for v in self.dthetaM], 'd'))))
-            if thetaM[-1] == 0: thetaM[-1] = 1
-            # if thetaM[-1] > 0:
+            if thetaM[-1] == 0:
+                thetaM[-1] = 1
+
             thetaM *= 1/thetaM[-1]
             PthetaM = clip(monospline(z, thetaM, Pz), 0, 1)
         else:
--- a/refl1d/magnetism.py
+++ b/refl1d/magnetism.py
@@ -480,7 +480,7 @@ class FreeMagnetismInterface(BaseMagnetism):
             tabove, name=name + " tabove", limits=(0, 360))
         if len(self.dz) != len(self.drhoM):
             raise ValueError("Need one dz for each drhoM")
-        if len(self.dthetaM) > 0 and len(self.drhoM) != len(self.dthetaM):
+        if 0 < len(self.dthetaM) != len(self.drhoM):
             raise ValueError("Need one dthetaM for each drhoM")
 
     def parameters(self):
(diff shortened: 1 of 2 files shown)


refl1d/magnetism.py
```diff
@@ -510,16 +510,16 @@ def profile(self, Pz, thickness):
         rhoM = hstack((0, cumsum(asarray([v.value for v in self.drhoM], 'd'))))
         # AJC added since without the line below FreeMagnetismInterface
         # does not initialise propoerly - fixes strange behaviour at drho=0 on end point
-        if rhoM[-1] == 0:
-            rhoM[-1] = 1
-        if rhoM[-1] > 0:
-            rhoM *= 1/rhoM[-1]
+        if rhoM[-1] == 0: rhoM[-1] = 1
```
pkienzle, line 513: Lint wants a line break after the colon. And similarly for thetaM

refl1d/magnetism.py
```diff
@@ -510,16 +510,16 @@ def profile(self, Pz, thickness):
...
+        # if rhoM[-1] > 0:
```
pkienzle, line 514: No need to keep this as a comment.

pkienzle (approved):
Looks fine. I didn't try running it.

pkienzle (approved):

acaruana2009:
I decided to add in the changes I discussed over email. This includes making thetaM (tabove or tbelow) the same if one or the other are not specified (this makes sure there is not an unintentional magnetic twist, say if the magnetic layer is interfaced with a non-magnetic layer.

refl1d/magnetism.py
```diff
@@ -505,21 +516,25 @@ def to_dict(self):
...
 
         rhoM = hstack((0, cumsum(asarray([v.value for v in self.drhoM], 'd'))))
         # AJC added since without the line below FreeMagnetismInterface
         # does not initialise propoerly - fixes strange behaviour at drho=0 on end point
         if rhoM[-1] == 0:
             rhoM[-1] = 1
-        if rhoM[-1] > 0:
-            rhoM *= 1/rhoM[-1]
+
+        rhoM *= 1/rhoM[-1]
```
pkienzle, line 529: The pre-existing variable names are misleading: `rhoM` and `thetaM` should be `rhoM_fraction` and `thetaM_fraction` or some such since the represent the left/right blend portion.

pkienzle (approved):
The code looks okay and I'm fine with it being merged (assuming it runs; I haven't tested!), but I suggested some changes.

I'm still not happy with the representation of freeform parameters. If you scale all dThetaM by a constant and you get the same solution once you normalize; this will confuse the fitter. At least one of the knots should be fixed, but which? This discussion is outside the scope of this PR.

refl1d/magnetism.py
```diff
@@ -474,13 +474,24 @@ def parvec(vector, name, limits):
...
+
+        # if only tabove or tbelow is defined then they are made equal
+        # this is to deal with the situation of a non-magnetic
+        # layer next to a magnetic one
+        if tabove is None and tbelow is None:
+            tbelow = tabove = DEFAULT_THETA_M
+        elif tabove is not None and tbelow is None:
+            tbelow = tabove
+        elif tbelow is not None and tabove is None:
+            tabove = tbelow
```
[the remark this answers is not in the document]
acaruana2009: I will adjust as suggested and push it through.

refl1d/magnetism.py
```diff
@@ -474,13 +474,24 @@ def parvec(vector, name, limits):
...
+
         self.tbelow = Parameter.default(
             tbelow, name=name + " tbelow", limits=(0, 360))
         self.tabove = Parameter.default(
             tabove, name=name + " tabove", limits=(0, 360))
         if len(self.dz) != len(self.drhoM):
             raise ValueError("Need one dz for each drhoM")
-        if len(self.dthetaM) > 0 and len(self.drhoM) != len(self.dthetaM):
+        if 0 < len(self.dthetaM) != len(self.drhoM):
```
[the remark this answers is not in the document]
acaruana2009: I think this was a result of my linter picking it up. I am happy to leave it as it was.

refl1d/magnetism.py
acaruana2009: Ok, I am happy to go through and change instances of `rhoM` and `thetaM` to `rhoM_fraction` and `thetaM_fraction` respectively.

acaruana2009:
Edits done as suggested. 

I agree fully about the FreeForm parameters representation. Is it worth opening an issue on it and discussing it there? Equally happy to discuss over email.

refl1d/magnetism.py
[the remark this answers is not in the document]
pkienzle: I'm surprised the linter complained. Unless it was complaining about the `len(a) > 0` clause?

Unlike lists, if `obj` is a numpy array then `if obj` raises an error rather than checking length so I habitually use `len(a) > 0` and ignore the linter.

I'm not averse to using new syntax, but a small amount of unscientific polling says the `0 < a != b` construct is harder to read.

pkienzle:
Discuss the freeform representation in a ticket.

acaruana2009:
Ok I have raised a ticket #112. I believe I have captured the problem you were mentioning, however, please if I have not covered the issue fully.

Issue #114, opened 2021-04-09
theta_offset is changing model calculation range


bmaranville:
For some reason when theta_offset is a applied, it is causing an error in the calculation of the model. 
With theta_offset=0:
![image](https://user-images.githubusercontent.com/686570/114211862-29e02480-992f-11eb-885c-349ca72589c4.png)
With nonzero theta_offset:
![image](https://user-images.githubusercontent.com/686570/114211951-45e3c600-992f-11eb-8449-36db0e5b75ab.png)

It is not just a visual artifact either, as the chisq is strongly affected.
[theta_offset_bug.zip](https://github.com/reflectometry/refl1d/files/6287218/theta_offset_bug.zip)

bmaranville:
This is with stable version Refl1D-0.8.14-exe.zip on Windows

bmaranville:
A proposed patch: don't know how it will affect performance
diff --git a/refl1d/probe.py b/refl1d/probe.py
index fd5e1d0..677828d 100644
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -1562,7 +1562,7 @@ def measurement_union(xs):
     TL = set()
     for x in xs:
         if x is not None:
-            TL |= set(zip(x.T, x.dT, x.L, x.dL))
+            TL |= set(zip(x.T+x.theta_offset.value, x.dT, x.L, x.dL))
     T, dT, L, dL = zip(*[item for item in TL])
     T, dT, L, dL = [np.asarray(v) for v in (T, dT, L, dL)]
 
@@ -1730,6 +1730,10 @@ class PolarizedNeutronProbe(object):
 
     @property
     def calc_Q(self):
+        #print('calculating calc_Q...')
+        self.T, self.dT, self.L, self.dL, self.Q, self.dQ \
+            = measurement_union(self.xs)
+        self._set_calc(self.T, self.L)
         return self.calc_Qo
 
     def _set_calc(self, T, L):

acaruana2009:
Hi,

I have taken a look at this and this occurs when you are using load4 as a list for the +ve theta_offset. If you just say load one data set per probe then it no longer does this. However, I am not convinced it is doing the right thing here either. Beyond a certain magnitude of theta offset the chi^2 returns NaN because the theory is no longer extending to the length of the data - additionally, I can recreate what you have shown with a -ve theta_offset for just data set per probe:
[...]

PR #116, opened 2021-04-13
Theta offset fix

bmaranville:
This fix should address the Q-range for PolarizedNeutronProbe not matching the Q-range for the cross-sections in the Probe, when a non-zero theta_offset is present in one of the cross-sections (#114)

Uses a the cached self.calc_Qo if the theta_offset values in the cross-sections do not change.
If all the theta_offset values are the same, it does a minimal calculation to update Q (this is a common situation)
If all the theta_offset values are not all the same, and they change, a full calculation of the probe Q the (measurement_union(xs)) is called.


bmaranville:
Tested, seems to show 10% slowdown in fits in worst case (single layer with microstepping disabled to emphasize model stepping overhead)

PR #118, opened 2021-04-22
ugly hack to support incompatibility between matplotlib and wx 

bmaranville:
FigureManager is calling window.SetTitle, where window is now the same thing as frame, and only top-level windows have SetTitle in new wx versions, not Panel

### DIFF
--- a/refl1d/view/model_view.py
+++ b/refl1d/view/model_view.py
@@ -95,6 +95,9 @@ class ModelView(wx.Panel):
         self._need_interactors = self._need_redraw = False
         self.Bind(wx.EVT_SHOW, self.OnShow)
 
+    def SetTitle(self, *args, **kw):
+        pass
+
     def OnContextMenu(self,event):
         """
         Forward the context menu invocation to profile, if profile exists.
--- a/refl1d/view/model_view.py
+++ b/refl1d/view/model_view.py
@@ -11,7 +11,6 @@ import numpy as np
 from matplotlib.figure import Figure
 from matplotlib.axes   import Subplot
 from matplotlib.backends.backend_wxagg import FigureCanvasWxAgg as FigureCanvas
-from matplotlib.backends.backend_wxagg import FigureManager
 from matplotlib.backends.backend_wxagg import NavigationToolbar2WxAgg
 
 from bumps.fitproblem import MultiFitProblem
@@ -50,9 +49,6 @@ class ModelView(wx.Panel):
         # Show toolbar or not?
         self.toolbar = NavigationToolbar2WxAgg( self.canvas )
         self.toolbar.Show(True)
-
-        # Create a figure manager to manage things
-        self.figmgr = FigureManager( self.canvas, 1, self )
 
         # Panel layout
         self.profile_selector_label = wx.StaticText(self, label="Sample")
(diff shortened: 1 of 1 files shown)


pkienzle:
Also needed in `refl1d.view.data_view.py` and `bumps.gui.plot_view.py`?

[Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) says this needs to be refactored as a plot panel mixin...

bmaranville:
It would be really nice if this would stop happening all the time.  wx and matplotlib do not play well together...

bmaranville:
It looks like FigureManagerBase is expecting to be run in a Frame, instead of a Panel.

bmaranville:
The mixin solution would require a release of bumps as well, since we build refl1d from the pypi bumps release.  Maybe I should split off the build for "unstable" so it uses the bumps "unstable" code as well, so we can fix things faster and get them to users.

pkienzle:
If we fed the embedding upstream then it might be better maintained. Or create an embedding package hosted separately on pip so others can help maintain it. I'm not sure that would be less work though.

bmaranville:
Actually I don't think this affects refl1d.view.data_view.py and bumps.gui.plot_view.py... they call FigureManagerBase through EmbeddedPylab, and not FIgureManager directly like model_view does. 

That might argue that model_view can be refactored to use EmbeddedPylab like the others do, but that would take quite a bit of time.

bmaranville:
Ok, I was looking at the code wondering if we could replace FigureManager with FigureManagerBase in model_view, but... it turms out that the FigureManager created in the __init__ is never used!  There is no other reference to 'figmgr' in that file, or any other file in bumps or refl1d.

I tried commenting out the line self.figmgr = FigureManager(...) line, and refl1d gui still loads.  Is there some side effect from instantiating FigureManager that is necesary, or can we just remove it?

pkienzle:
`FigureManager(canvas)` sets `canvas.manager` so it is doing something. It'll require quite a bit of checking to make sure all the figure interactions still work. For example, can you still pan and zoom in the profile window?

bmaranville:
Zooming and panning in the profile view seems to work fine still.  What other interactions should I check?

bmaranville:
Oh - right - all the interactors!  Trying them out now.

bmaranville:
It seems to work fine, though if you'd rather we could just switch FigureManagerBase for FIgureManager and it would fix the conflict, at least for now.  Setting canvas.manager happens in FigureManagerBase.

FigureManagerBase calls self.set_title now, as of version 3.4.1 of matplotlib, but it does *not* implement set_title (just a stub) so it has no effect.  FigureManagerWx has set_title implemented, and calls self.window.SetTitle(), which is what triggered the bug.

pkienzle:
Looking for `.manager` in matplotlib:
* `animation.py` - don't care
* `backend_bases.py` - for `full_screen_toggle()` but properly trapped
* `figure.py` - for `figure.show()` which we don't use.
* `pyplot.py` - for `figure()` and `get_current_figure_manager()` which we don't use.
* `widgets.py` - don't care

So it should be fine commented out, but we might have better future-proofing with `FigureManagerBase`.

bmaranville:
Ok, I'll use FigureManagerBase and test.

bmaranville:
Seems to work.  I think we can jettison this pull request and just make a two-line fix in master.

Apparently we can't convert a pull request to an issue (though the reverse is possible).  Should we keep it for posterity, or dump?

glass-ships:
Is this branch safe to delete, if it's been implemented elsewhere?

PR #126, opened 2021-06-08
Suggested fix for theta_offset problem in PolarizedNeutronProbe

bmaranville:
The optimization of just adding a fixed offset when all the theta_offsets are shared
in a PNP object does not work, particularly if the data needs to be oversampled.

This commit removes the optimized version which doesn't work, and instead adds an
"oversampling" argument to the init of PolarizedNeutronProbe.

If the theta_offsets change (if they are a fit parameter) the measurement_union and
the resampling are recalculated.  This is slow but there may be no way around this
without a mesh of theta/lambda values to interpolate from.

### DIFF
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -1238,7 +1238,7 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
           L=None, dL=None, T=None, dT=None, dR=None,
           FWHM=False, radiation=None,
           columns=None, data_range=(None, None),
-          resolution='normal',
+          resolution='normal',oversampling=None,
          ):
     r"""
     Load in four column data Q, R, dR, dQ.
@@ -1327,6 +1327,9 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
 
     *resolution* is 'normal' (default) or 'uniform'. Use uniform if you
     are merging Q points from a finely stepped energy sensitive measurement.
+
+    *oversampling* is None or a positive integer indicating how many points to add
+    between data point to support sparse data with denser theory (for PolarizedNeutronProbe)
     """
     entries = parse_multi(filename, keysep=keysep, sep=sep, comment=comment)
     if columns:
@@ -1368,7 +1371,7 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
         if any(isinstance(d, QProbe) for d in xs if d is not None):
             probe = PolarizedQProbe(xs, Aguide=Aguide, H=H)
         else:
-            probe = PolarizedNeutronProbe(xs, Aguide=Aguide, H=H)
+            probe = PolarizedNeutronProbe(xs, Aguide=Aguide, H=H, oversampling=oversampling)
     return probe
 
 def _data_as_probe(entry, probe_args, T, L, dT, dL, dR, FWHM, radiation,
@@ -1609,9 +1612,10 @@ class PolarizedNeutronProbe(object):
     show_resolution = None  # Default to Probe.show_resolution when None
     substrate = surface = None
     polarized = True
-    def __init__(self, xs=None, name=None, Aguide=BASE_GUIDE_ANGLE, H=0):
+    def __init__(self, xs=None, name=None, Aguide=BASE_GUIDE_ANGLE, H=0, oversampling=None):
         self._xs = xs
         self._theta_offsets = None
+        self.oversampling = oversampling
 
         if name is None and self.xs[0] is not None:
             name = self.xs[0].name
@@ -1724,7 +1728,7 @@ class PolarizedNeutronProbe(object):
                 x.theta_offset = theta_offset
                 x.sample_broadening = sample_broadening
 
-    def oversample(self, n=6, seed=1):
+    def _oversample(self, n=6, seed=1):
         # doc string is inherited from parent (see below)
         rng = numpy.random.RandomState(seed=seed)
         T = rng.normal(self.T[:, None], self.dT[:, None], size=(len(self.dT), n))
@@ -1732,29 +1736,27 @@ class PolarizedNeutronProbe(object):
         T = T.flatten()
         L = L.flatten()
         self._set_calc(T, L)
-    oversample.__doc__ = Probe.oversample.__doc__
+    _oversample.__doc__ = Probe.oversample.__doc__
 
     def _calculate_union(self):
         theta_offsets = [x.theta_offset.value for x in self.xs if x is not None]
+        # print('theta_offsets', self._theta_offsets, theta_offsets)
         if self._theta_offsets is not None and theta_offsets == self._theta_offsets:
-            # no change in offsets: use cached values of measurement union
-            return
+           # print("no offset change... returning", self._theta_offsets, theta_offsets)
+           # no change in offsets: use cached values of measurement union
+           return
 
-        unique_offsets = set(theta_offsets)
-        shared_offset = theta_offsets[0] if len(unique_offsets) == 1 else None
-        if self._theta_offsets is not None \
-            and shared_offset is not None \
-            and len(set(self._theta_offsets)) == 1:
-            # offset was shared, and is shared, but value changed
-            Q = TL2Q(T=self.T + shared_offset, L=self.L)
-            self.calc_Qo = Q
         else:
             # unshared offsets changed, or union has not been calculated before
             self.T, self.dT, self.L, self.dL, self.Q, self.dQ \
                 = measurement_union(self.xs)
-            self._set_calc(self.T, self.L)
 
-        self._theta_offsets = theta_offsets
+            if self.oversampling is None:
+                self._set_calc(self.T, self.L)
+            else:
+                self._oversample(self.oversampling)
+            
+            self._theta_offsets = theta_offsets
 
     @property
     def calc_Q(self):


bmaranville:
It seems like the intent of the method PolarizedNeutronProbe.oversample is to have a persistent oversampling... I can rework the interface so that it is applied the same was as before, but the _oversample method is called after every time the measurement_union is recalculated (if the theta_offset values of any of the cross sections change).  We can document for the users that fitting theta_offset in a Polarized probe with oversampling will lead to some slowdown.

bmaranville:
But there are issues with the "critical_edge" method, too - which is also supposed to be persistent.  This is a complicated problem.

pkienzle:
The interface has changed so a test is breaking: `doc/examples/freemag/pmf.py`. Easy enough to fix as you suggest above.

It is a little disturbing that changing the theta offset for one cross section can change the computed values for the other cross sections. That's because the set of Q values returned by measurement union may change, which will change the input to the convolution used for each data point. Also the number of calculated Q points will jump by a factor of four if a system with nominally the same angles for each cross section has independent theta offset.

The situation is worse for oversampling. Here there is change the order of calls to the random number generator which will change the cloud of calculated points generated around each data point.  It may be slightly better to put the point clouds around each point before the measurement union (sharing clouds when theta matches) and apply the theta offset before the measurement union. However, we would still see the 4x number of Q values in the measurement union, and the changed result from more points in the convolution so not a big improvement.

Instead of changing the calculated Q values, we may want to keep them fixed and rely on the convolution to sort thing out. The data Q will still change according to theta offset but the convolution will treat the calculated theory as a linear spline when evaluating the theory at that Q, so values will remain approximately correct. For theta offset much smaller than resolution this should be fine. If theta offset gets too large we will run into problems at the end points where calculated Q values cannot support the resolution calculation. This can be addressed by providing a bit of padding for the calculated Q at the ends, which is something we should be doing regardless.

A middle ground is to interpolate the theory from the calculated Q into adjusted data (possibly with oversampling) before convolution. This allows us to use a monotonic spline to follow the reflectivity curve rather than the simple linear spline used in the resolution calculation.

bmaranville:
If you think extending the ends is a viable solution, that seems easier.  That's how we got to this in the first place, as large theta_offset values were leading to zeros in the theory function near the endpoint, giving large chisq values (it seemed to be failing in a bad way, which is perhaps a separate ticket).  Users discovered this in unstitched TOF data where the endpoints were at low Q, so the effects were pretty dramatic.

________________________________

pkienzle:
The lack of extension beyond the ends in an outstanding issue from before the first public release. I was testing the effects of cutoff in the resolution and found that with proper normalization it worked out okay and I never got around to extending beyond the ends like mlayer does.

The correct approach is to make sure every q point has a basis of support spanning at least 3 ΔQ, including all middle points and end points.  And a little wider if we want to allow theta offset effects. Furthermore, if we add in an overall thickness estimate, we can make sure we oversample appropriately to cover all the fringes. But for now I would try just extending the ends, and hope not too many datasets are excluding points in the middle.

bmaranville:
After thinking about it some more, I'm not sure how to sensibly implement the extension of the endpoints.  How much do you extend it?  What is the scale of "reasonable"?  Should we pull in the range of the theta_offset parameter?

bmaranville:
Ok - I think I have an idea of how to implement an extension of the endpoints.  

The strategy I'm thinking will work based on what Paul recommended is:

- get range [TO_lower, TO_upper] of theta_offset (TO) for each cross-section (intersection of limits and user bounds)
- generate a list of theta values (T) for each cross section that is made of these three segments concatenated:
   - from max(0, Tmin - TO_lower) to Tmin in steps of dT == T[1] - T[0] (the first step in the original T)
   - the original list of T
   - from T_max to T_max + TO_upper in steps of the same size as the last step in the original T

From there I can generate the measurement union, and oversampling only has to be done once.  There is still a complication that we somehow have to trigger a recalculation if the limits of the theta_offset ever change.

bmaranville:
Ok, the new version has been pushed here.  It only recalculates the measurement union if the limits of the theta_offset Parameter are changed.  
I think it's ready for review.

refl1d/probe.py
```diff
@@ -1712,7 +1750,8 @@ def shared_beam(self, intensity=1, background=0,
         back_absorption = Parameter.default(back_absorption,
                                             name="back_absorption",
                                             limits=[0, 1])
-        theta_offset = Parameter.default(theta_offset, name="theta_offset")
+        theta_offset = Parameter.default(theta_offset,
+                                            name="theta_offset")
```
pkienzle, line 1754: Optional: I'm leaning toward black's strategy of breaking after the parenthesis for long argument lists. This leads to a consistent four character indent on every line rather than trying to line up the indent with the parenthesis.
So:
```
theta_offset = Parameter.default(
    theta_offset, name="theta_offset")
```
Black may be more verbose about it for consistency:
```
theta_offset = Parameter.default(
    theta_offset,
    name="theta_offset",
)
```
but I prefer to the more compact vertical form.

refl1d/probe.py
```diff
@@ -1725,36 +1764,44 @@ def shared_beam(self, intensity=1, background=0,
...
+    def _oversample(self, n=6, seed=1):
         # doc string is inherited from parent (see below)
         rng = numpy.random.RandomState(seed=seed)
         T = rng.normal(self.T[:, None], self.dT[:, None], size=(len(self.dT), n))
         L = rng.normal(self.L[:, None], self.dL[:, None], size=(len(self.dL), n))
         T = T.flatten()
         L = L.flatten()
         self._set_calc(T, L)
-    oversample.__doc__ = Probe.oversample.__doc__
+    _oversample.__doc__ = Probe.oversample.__doc__
```
pkienzle, line 1779: doc string belongs with `oversample` instead of `_oversample`.

refl1d/probe.py
```diff
@@ -1725,36 +1764,44 @@ def shared_beam(self, intensity=1, background=0,
...
-            self.calc_Qo = Q
+        theta_offset_limits = []
+        for xs in self.xs:
+            if xs is None or xs.theta_offset.fixed:
+                theta_offset_limits.append(None)
+            else:
+                theta_offset_limits.append(xs.theta_offset.bounds.limits)
+
+        if self._theta_offset_limits is not None and theta_offset_limits == self._theta_offset_limits:
+           # no change in offset limits: use cached values of measurement union
```
pkienzle, line 1790: The current code uses `theta_offset` value when calculating measurement union, so you would need to check for that changing as well.

refl1d/probe.py
```diff
@@ -1725,36 +1764,44 @@ def shared_beam(self, intensity=1, background=0,
...
                 = measurement_union(self.xs)
-            self._set_calc(self.T, self.L)
 
-        self._theta_offsets = theta_offsets
+            if self.oversampling is None:
+                self._set_calc(self.T, self.L)
+            else:
+                self._oversample(self.oversampling, self.oversampling_seed)
+            
+            self._theta_offset_limits = theta_offset_limits
```
pkienzle, line 1804: Can dedent and remove `else` since `if` ends in `return`.

refl1d/probe.py
bmaranville: My logic was that if it was fixed, the value wouldn't be changing, so that you don't need to explicitly check the value.  Is there a situation where this will fail?

pkienzle: Yes, it seems that the code should work like you expect. I just saw that `theta_offset.value` was being referenced; I didn't actually look at how it was being used.

pkienzle: Okay, here's a scenario: In the GUI, fix theta offset range and change its value manually. It is still fixed, so no triggering of the recalc, but the Q_calc list will be different when you save and reload.

bmaranville: Good catch!

refl1d/probe.py
```diff
@@ -1561,12 +1565,44 @@ def measurement_union(xs):
...
+            dT_upper_extension = np.ones_like(T_upper_extension, dtype=dT.dtype) * dT[-1]
+            L_upper_extension = np.ones_like(T_upper_extension, dtype=L.dtype) * L[-1]
+            dL_upper_extension = np.ones_like(T_upper_extension, dtype=dL.dtype) * dL[-1]
+
+            T = np.hstack([T_lower_extension, T, T_upper_extension])
+            dT = np.hstack([dT_lower_extension, dT, dT_upper_extension])
+            L = np.hstack([L_lower_extension, L, L_upper_extension])
+            dL = np.hstack([dL_lower_extension, dL, dL_upper_extension])
+
+            TL |= set(zip(T, dT, L, dL))
```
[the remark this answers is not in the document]
bmaranville: Should we alter the __init__ for PolarizedNeutronProbe so that it checks for a fixed geometry across all cross sections, then store only the one?

refl1d/probe.py
```diff
@@ -1368,7 +1372,7 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
         if any(isinstance(d, QProbe) for d in xs if d is not None):
             probe = PolarizedQProbe(xs, Aguide=Aguide, H=H)
         else:
-            probe = PolarizedNeutronProbe(xs, Aguide=Aguide, H=H)
+            probe = PolarizedNeutronProbe(xs, Aguide=Aguide, H=H, oversampling=oversampling)
```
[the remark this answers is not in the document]
bmaranville: In the current form, it does accept a seed which is stored as an instance attribute - I was wondering, though: why does e.g. QProbe.oversample have a default seed value of 1 instead of None?  Woudn't it be preferable to shake the dice if the user doesn't specify a seed value explicitly?

refl1d/probe.py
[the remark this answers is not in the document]
bmaranville: If we oversample and extend in Q-space, we will not be able to support PolarizedNeutronProbe.unique_L and PolarizedNeutronProbe._L_idx attributes, which are calculated from an explicitly oversampled L, and there is no unique way to go back from Q -> T,L  ... so supporting wavelength-dependent SLD will get hard, will it not?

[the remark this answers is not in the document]
bmaranville: I guess the code is already not oversampling in L properly, anyway, since we are currently sorting on T and then oversampling L in T order.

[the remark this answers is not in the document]
pkienzle: I think we need to create a θ-λ grid for the energy dependent case, with separate oversampling in θ and λ. Render the microslabs and calculate the reflectivity for each λ separately, then integrate over θ-λ patches for the resolution.

I had originally planned to allow a separate ρ for each q, but hadn't worked out the details of how to generate them.

Anyway, we could drop the L_idx attribute since it was never used.

bmaranville:
Latest commit is a sort-of-working implementation of extending in Q instead of Theta.  It is a bunch of work to try to untangle all the wavelength stuff (unique_L has to be disabled) and the tests are failing because self.L is not defined properly still.  I don't really like this approach and am still not convinced it's worth the extra complexity compared to recalculating the measurement union whenever theta_offset changes, as is done in 9ce32df47900f4f2608156db74771156f8d9189a - for problems that don't use oversampling, that version is faster.  For problems that do use oversampling, it's only minimally slower.

In quick checks with real models, I don't see any performance gains from extending the range compared to recalculating the measurement_union at each step.  I think the extra Q values for which you have to then calculate R make it a wash performance wise.

I'd be happy to revert to 9ce32df47900f4f2608156db74771156f8d9189a and call it done.

bmaranville:
I think this is ready for review.  I converted the critical_edge method into a state-parameter setter like for oversampling, so that they can both be applied simultaneously (nearly: oversampling is always applied first) when the q-basis is recalculated, and the q-basis is recalculated every time anyone touches: (theta_offset limits, theta_offset values if fixed, oversampling params, critical_edge_oversampling params) but left alone otherwise.

The same logic is applied to Probe and PolarizedNeutronProbe for the discussion above.  QProbe and ProbeSet have not been modified.

refl1d/probe.py
```diff
@@ -180,7 +239,9 @@ def __init__(self, T=None, dT=0, L=None, dL=0, data=None,
                  intensity=1, background=0, back_absorption=1, theta_offset=0,
                  sample_broadening=0,
                  back_reflectivity=False, name=None, filename=None,
-                 dQ=None, resolution='normal'):
+                 dQ=None, resolution='normal',
+                 oversampling: Optional[Union[float, Oversampling]] = None,
+                 critical_edge_oversampling: Optional[CriticalEdgeOversampling] = None):
```
pkienzle, line 244: Need to add oversampling/critical_edge_oversampling to the doc strings.

I'm not too keen on the interface for oversampling:
```
Probe(..., oversampling={'n': 5, 'seed': 12}, ...)
```

Shall we drop the sampling and just use a simple `n=...` selected linearly between the neighbouring points? Or maybe set some maximum δQ out and make sure there is coverage out to ±3ΔQ? Instead of δQ maybe make it a boolean and set δQ = 2π/5w where w is the total thickness of the sample? Instead of specifying oversampling, look at the theory function and apply it if the sample is thick and the resolution is bigger than δQ?

Similarly for critical edge oversampling `substrate` and `surface` can come from the model, `n` can come from the model thickness and `delta` can come from the resolution. To get the info we need from the model change Q_calc to a function which takes these details as parameters.

Probably more work than you have time for right now.

In lieu of a rewrite, it might be nice to allow the user to hand over the substrate/surface materials rather than making these fixed constants.

bmaranville: I think going the other direction might work even better... Just have the user specify the Qc, which they can read right off a plot?

glass-ships:
Is this branch something we plan on reviving?

bmaranville:
maybe.  Oversampling is expensive and this approach works pretty well under some circumstances.

PR #127, opened 2021-06-10
FIX: Change errors.py calc_errors() to produce a smooth magnetic profile

acaruana2009:
Currently the contours generated for magnetic profiles for the non-sliced samples are producing a step profile. Given the fact that we are already using smooth_profile() for non-magnetic profiles, we should change this to be consistent.


bmaranville (commented):
I think that the microslabber is doing a "blend" operation on thetaM now, which is closer to what the Nevot-Croce approximation corresponds to - so there's no reason not to use the smooth profile representation (a long time ago there might have been an inconsistency in what was represented)

pkienzle:
Yes, that makes sense, both why it is and why it should change. Feel free to pull.

PR #130, opened 2021-07-08
look for resolution_shape='normal|uniform' in data file header

pkienzle:


bmaranville:
@pkienzle - did you check that this does grab the resolution from a file when it is specified therein?  If so, I have no objections to this and will merge immediately.

pkienzle:
I haven't tested. https://github.com/reductus/reductus/pull/130 implements the `resolution_shape` header key but it isn't merged yet.

Is that what we want to call it? Will we have the same shape for each data point? Do we want to jump immediately to a more general ORSO scheme?

bmaranville:
It seems like there are still some unanswered questions around this.  For one, I wonder if we need to be clear that the 'normal' resolution function is truncated (and that the truncation is 3-sigma?).  My vote is to defer merging until after the upcoming release.

pkienzle:
There's an easy workaround so I have no problem with a delay. May want to ask some actual users though.

glass-ships:
Been a few years - should we revisit this?

bmaranville:
Yes, it needs to be manually recreated since `load4` has changed some and we also want to support reading this metadata from orso files.

pkienzle:
The patch will no longer apply. Leave this as a ticket.

PR #131, opened 2021-07-09
Magnetic reflectivity py

bmaranville:
This was an interesting experiment... I copied our magnetic_amplitude code from C back into python (almost verbatim) and compiled it with numba (Ahead Of Time compile)

The resulting calculations of a real model are around 20% faster with numba than the C extension: (the first number is for the C extension, the second for the numba AOT extension)

<pre>(refl1d-3.8) <font color="#8AE234"><b>brian@cheery</b></font>:<font color="#729FCF"><b>~/dev/refl1d-modelbuilder/helical_simulation</b></font>$ time refl1d  ~/Downloads/Combined_Simple_2Gap_Top.txt --store=path --time_model --steps=1000
time per model eval: 41.836 ms

real	0m43.111s
user	0m43.218s
sys	0m0.314s
(refl1d-3.8) <font color="#8AE234"><b>brian@cheery</b></font>:<font color="#729FCF"><b>~/dev/refl1d-modelbuilder/helical_simulation</b></font>$ time refl1d  ~/Downloads/Combined_Simple_2Gap_Top.txt --store=path --time_model --steps=1000
loaded from compiled module
time per model eval: 34.6037 ms

real	0m36.062s
user	0m36.163s
sys	0m0.335s
</pre>

The code is not really ready to use - the importing or compiling of the module dance is a bit awkward as written.


refl1d/reflectivity.py
```diff
@@ -115,14 +117,13 @@ def reflectivity_amplitude(kz=None,
...
     irho[irho < 0] = 0.
-    #print depth.shape, rho.shape, irho.shape, sigma.shape
-    #print depth.dtype, rho.dtype, irho.dtype, sigma.dtype
+    # print depth.shape, rho.shape, irho.shape, sigma.shape
+    # print depth.dtype, rho.dtype, irho.dtype, sigma.dtype
     r = np.empty(kz.shape, 'D')
-    #print "amplitude", depth, rho, kz, rho_index
-    #print depth.shape, sigma.shape, rho.shape, irho.shape, kz.shape
+    # print "amplitude", depth, rho, kz, rho_index
+    # print depth.shape, sigma.shape, rho.shape, irho.shape, kz.shape
```
pkienzle, line 126: Personal style: I use no spaces when commenting out code vs. one space when writing comments about code. *Black* will of course ignore me and put a space in there.

refl1d/reflectivity.py
```diff
@@ -213,20 +217,26 @@ def magnetic_amplitude(kz,
...
 
-    #kz = -kz
-    #depth, rho, irho, sigma, rhoM, thetaM = [v[::-1] for v in (depth, rho, irho, sigma, rhoM, thetaM)]
-    depth, rho, irho, sigma = [_dense(a, 'd') for a in (depth, rho, irho, sigma)]
-    #np.set_printoptions(linewidth=1000)
-    #print(np.vstack((depth, np.hstack((sigma, np.nan)), rho, irho, rhoM, thetaM)).T)
+    # kz = -kz
+    # depth, rho, irho, sigma, rhoM, thetaM = [v[::-1] for v in (depth, rho, irho, sigma, rhoM, thetaM)]
+    depth, rho, irho, sigma = [_dense(a, 'd')
+                                      for a in (depth, rho, irho, sigma)]
```
pkienzle, line 223: Maybe indent as:
```
LHS = [
    list contents]
```
or
```
LHS = [
    list contents,
]
```
I find this makes for less work when changing LHS. You don't need to reindent the continuation lines.

refl1d/reflectivity.py
```diff
@@ -238,33 +248,36 @@ def calculate_u1_u3(H, rhoM, thetaM, Aguide):
...
     # the direction of the magnetic field outside the sample.
     if rotate_M:
         # rotate the M vector instead of the transfer matrix!
         # First, rotate the M vector about the x axis:
-        new_my = sld_m_z * sin(radians(Aguide)) + sld_m_y * cos(radians(Aguide))
-        new_mz = sld_m_z * cos(radians(Aguide)) - sld_m_y * sin(radians(Aguide))
+        new_my = sld_m_z * sin(radians(Aguide)) + \
+                               sld_m_y * cos(radians(Aguide))
+        new_mz = sld_m_z * cos(radians(Aguide)) - \
+                               sld_m_y * sin(radians(Aguide))
```
pkienzle, line 273: PEP8 discourages the use of backslash for line extension. Use parens instead.  And break before the operator.
```
new_my = (sld_m_z * sin(radians(Aguide))
    + sld_m_y * cos(radians(Aguide))
```
I'm guessing *Black* would do:
```
new_my = (
    sld_m_z * sin(radians(Aguide))
    + sld_m_y * cos(radians(Aguide)
)
```
I prefer fewer vertical lines, but the other version is defensible.

refl1d/reflectivity.py
```diff
@@ -285,33 +298,340 @@ def calculate_u1_u3_py(H, rhoM, thetaM, Aguide):
...
+
+
+    #    Calculate reflectivity coefficients specified by POLSTAT
+    YA[i] = (B24*B41-B21*B44)/DETW # ++
+    YB[i] = (B21*B42-B41*B22)/DETW # +-
+    YC[i] = (B24*B43-B23*B44)/DETW # -+
+    YD[i] = (B23*B42-B43*B22)/DETW # --
+
+@cc.export('mag_amplitude', 'void(f8[:], f8[:], f8[:], f8[:], f8[:], c16[:], c16[:], f8, f8[:], i4[:], c16[:], c16[:], c16[:], c16[:])')
+@njit(parallel=True, cache=True)
```
pkienzle, line 578: Put the type signature with the `njit` rather than the `cc`.

refl1d/reflectivity.py
```diff
@@ -285,33 +298,340 @@ def calculate_u1_u3_py(H, rhoM, thetaM, Aguide):
...
         # Otherwise we have @njit(...), so return the identity decorator.
         return lambda fn: fn
 
-@njit('(f8[:], f8[:], f8[:], f8[:], f8[:])')
+MINIMAL_RHO_M = 1e-2  # in units of 1e-6/A^2
+EPS = np.finfo(float).eps
+B2SLD = 2.31604654  # Scattering factor for B field 1e-6
+
+from numba.pycc import CC
+cc = CC('magnetic_amplitude')
```
pkienzle, line 326: Any reason to do AOT compile rather than simply shipping numba and letting it use whatever optimization is available for the platform it is running on? Sounds like less work to me. Startup times will be slightly slower, but not an issue when fitting since the compile only needs to happen once. Or maybe once per process: I don't know how numba interacts with multiprocessing.

Note that numba is able to parallelize across cores and sockets on the same machine, so it may be faster to run bumps single threaded with the nllf running in parallel rather than running the optimizer in parallel with the nllf single threaded. Or maybe fewer threads than cores to allow some overlap during the non-parallel bits while reducing contention during the parallel bits.

refl1d/reflectivity.py
```diff
@@ -452,63 +772,67 @@ def convolve_sampled(xi, yi, xp, yp, x, dx):
...
-    _check_uniform("uniform wide", xi, yi, x=[2.5, 4.5, 6.5, 8.5], dx=3)
+    xi=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+    yi=[1, 3, 1, 2, 7, 3, 1, 2, 1, 3]
+    _check_uniform("uniform aligned", xi, yi, x = [2, 4, 6, 8], dx = 1)
+    _check_uniform("uniform unaligned", xi, yi,
+                   x = [2.5, 4.5, 6.5, 8.5], dx = 1)
+    _check_uniform("uniform wide", xi, yi, x = [2.5, 4.5, 6.5, 8.5], dx = 3)
     # Check bad values
-    ystar = convolve(xi, yi, x=[-3, 13], dx=[1/np.sqrt(3)]*2, resolution='uniform')
+    ystar = convolve(xi, yi, x = [-3, 13], dx = [1/np.sqrt(3)]*2, resolution ='uniform')
```
pkienzle, line 782: PEP8: Don't use spaces around the = sign when used to indicate a keyword argument.

refl1d/reflectivity.py
```diff
@@ -452,63 +772,67 @@ def convolve_sampled(xi, yi, xp, yp, x, dx):
...
 def _check_uniform(name, xi, yi, x, dx):
     # Note: using fixed dx since that's all _check_spline supports.
-    ystar = convolve(xi, yi, x, [dx/np.sqrt(3)]*len(x), resolution='uniform')
-    #print("xi", xi)
-    #print("yi", yi)
-    #print("x", x, "dx", dx)
-    #print("ystar", ystar)
-    xp = [-dx, -dx, dx, dx]
-    yp = [0, 0.5/dx, 0.5/dx, 0]
+    ystar=convolve(xi, yi, x, [dx/np.sqrt(3)]*len(x), resolution = 'uniform')
```
pkienzle, line 792: PEP8: Always surround these binary operators with a single space on either side: assignment (=), augmented assignment (+=, -= etc.), comparisons (==, <, >, !=, <>, <=, >=, in, not in, is, is not), Booleans (and, or, not).

refl1d/reflectivity.py
bmaranville: The startup times were pretty long when using njit without AOT... 10+ seconds on my machine.  There are probably things that can be done to speed this up.  I found that for whatever reason the njit code was slightly slower than the AOT code, though that's not supposed to be true.

refl1d/reflectivity.py
bmaranville: I wasn't aware that this would work!

refl1d/reflectivity.py
bmaranville: I think VS code did that to me.  I don't know why it's not following PEP8

refl1d/reflectivity.py
pkienzle: Caching should take care of that. You are doing this by hand anyway, so let numba take care of it. It seems to work for me. It took 36s to run check_examples.py with njit alone and 38s to run with precompiled, in both cases having updated the cache prior to timing.

bmaranville: I found that it doesn't cache between starts of refl1d - it recompiles every time the program is started, which is potentially very annoying.

pkienzle: That hasn't been my experience (mac conda, python=3.8, numba=0.53.1). If you touch the file it'll recompile, but otherwise it should use the cache. Any chance you are touching the .py file between runs? [obviously not, but I have to ask...]

There are some issues on [this](https://github.com/numba/numba/issues/2439) from [two years ago](https://stackoverflow.com/questions/56936573/caching-jit-compiled-functions-in-numba). Perhaps they weren't resolved?

Are there any warnings showing up if you run from the command line?

The [docs](https://numba.pydata.org/numba-doc/latest/developer/caching.html) tell you how to set the cache directory. Can you check that the cache is being created?

bmaranville:
I had to fix up zeeman_check.py, but the new numba calculation seems to check out compared to the canonical gepore result:
image

pkienzle:
Sometimes shorter code can help the compiler and processor. I saved a little time (5%) by simplifying the B initialization, using the following:
```python
    Z = 0.
    S1LP = -sqrt(complex(PI4*(RHO[L]+RHOM[L]) - E0, -PI4*(fabs(IRHO[L])+EPS)))
    S3LP = -sqrt(complex(PI4*(RHO[L]-RHOM[L]) - E0, -PI4*(fabs(IRHO[L])+EPS)))
    BLP = GLP = 0j
    B11 = B22 = B33 = B44 = 1. + 0j
    B12 = B13 = B14 = B21 = B23 = B24 = B31 = B32 = B34 = B41 = B42 = B43 = 0j
    for I in range(0,N-1):
        ...
```
Note: I didn't check that the result was correct; I only wanted to see if it would be faster.

This may be worth doing to make the code easier to maintain. Fewer repeated expressions. I'll let you decide.

pkienzle:
Please change `dtype='complex128'` to `dtype=np.complex128`. Numba was complaining on one of my environments when it uses the string version. We may want to try returning a tuple rather than allocating and passing in a vector for Y.

Or we can return a tuple instead of passing in a numpy output array:
```python
CR4XA_SIG = 'UniTuple(c16,4)(i8, f8[:], f8[:], f8, f8[:], f8[:], f8[:], c16[:], c16[:], f8)'
```

Then use code like the following to receive it:
```
Ra[i], Rb[i], _, _ = Cr4xa(...)
```
etc.

No difference in performance.

refl1d/reflectivity.py
bmaranville: You are correct - the cache seems to be working.  I think I was thrown off because I was editing some other parts of the source file (not in the numba functions) and I didn't think that would affect the cache, but it does.

pkienzle:
I seem to be in micro-optimization mode.

It's probably too small to notice, but the following should be slightly faster:
```python
MAGAMP_SIG = 'c16[:,:](f8[:], f8[:], f8[:], f8[:], f8[:], c16[:], c16[:], f8[:])'
@njit(MAGAMP_SIG, parallel=True, cache=True)
def magnetic_amplitude_py(d, sigma, rho, irho, rhoM, u1, u3, KZ):
    R = np.empty((len(points), 4), dtype=np.complex128)
    for i in prange(points):
        Cr4xa(layers, d, sigma, 1.0, rho, irho, rhoM, u1, u3, KZ[i], R[i])
    # If magnetism in incident or backing media then recompute with IP=-1.0
    # assigning to R[2] and R[3] only. This must come after the calculation for IP=1.0
    # which assigns to all R[0], R[1], R[2], R[3].
    if abs(rhoM[0]) <= MINIMAL_RHO_M and abs(rhoM[layers-1]) <= MINIMAL_RHO_M:
        for i in prange(points):
            Cr4xa(layers, d, sigma, -1.0, rho, irho, rhoM, u1, u3, KZ[i], R[i])
    return R
```
modifying CR4XA to only assign `Y[0]`, `Y[1]` when IP > 0.

This avoids memory thrashing and copying associated with Y, allocating the entire return vector in one step.

bmaranville:
> Sometimes shorter code can help the compiler and processor. I saved a little time (5%) by simplifying the B initialization, using the following:
> 
> ```python
>     Z = 0.
>     S1LP = -sqrt(complex(PI4*(RHO[L]+RHOM[L]) - E0, -PI4*(fabs(IRHO[L])+EPS)))
>     S3LP = -sqrt(complex(PI4*(RHO[L]-RHOM[L]) - E0, -PI4*(fabs(IRHO[L])+EPS)))
>     BLP = GLP = 0j
>     B11 = B22 = B33 = B44 = 1. + 0j
>     B12 = B13 = B14 = B21 = B23 = B24 = B31 = B32 = B34 = B41 = B42 = B43 = 0j
>     for I in range(0,N-1):
>         ...
> ```
> 
> Note: I didn't check that the result was correct; I only wanted to see if it would be faster.
> 
> This may be worth doing to make the code easier to maintain. Fewer repeated expressions. I'll let you decide.

Sure!  Do we have a good test for magnetic reflectivity calculations, besides zeeman_check (which is not set up to be a CI test, as far as I can tell)?

pkienzle:
I run the sample models before releasing but only as a smoke test; I don't look at the results.

This is done using `python check_examples.py --chisq`.

It wouldn't take too much to turn this into a CI test. Just check if the chisq_str has changed. Meanwhile, you can do this manually using `python run doc/examples/spinvalve/n101G.py --chisq` before and after code mods.

bmaranville:
The numba convolve gaussian seems to perform almost exactly the same as the (nearly identical) C code from which it was copied.  Not very surprising.

bmaranville:
Ok, once I stopped treating c pointer arithmetic as numpy vector math (whoops) everything behaves much better.  The test cases give the same chisq as they did before, and they run 15-20% faster!

pkienzle:
You can try [fastmath](https://numba.pydata.org/numba-doc/latest/user/performance-tips.html#fastmath) in the jit call. Just set `fastmath=True`. Maybe it'll help with the `y += expr`, but probably not. Anyway, the gains will be incremental if any.

For fixed theta offset and sample broadening you see a lot more improvement using a sparse matrix multiply with pre-computed coefficients. You may even be able to get away with some theta offset (the relative q values aren't changing very much, so the result won't change either). That's not the case with sample broadening, and I doubt there is a simple tweak you can do to the coefficients to fix it. Or maybe…

Consider that the coefficients represent particular q points for a given Δq. If you change Δq, then you can reuse the coefficients if you can move the q points. But moving q means changing from I(q) to I(q'). So long as the change to Δθ is small, then the required q' will be near to q, and you should be able to approximate it with interpolation.  You might be able to do this with a matrix multiply, especially if you tweak the `q_calc` a little before computing `I(q_calc)`. I'm not sure that will be faster than recomputing the coefficients.

bmaranville:
I think if we can show that a numba kernel will be as fast or faster than the C kernel, that will be a big win (and so far, all my testing seems to indicate that it is a little bit faster).  Smaller optimizations can follow, but the fact that we could distribute a pure python package would be great from a maintenance perspective.

bmaranville:
@pkienzle this branch now runs and passes all tests without compiling reflmodule.  
These have all been ported to numba:

- convolve, convolve_uniform, convolve_gaussian, convolve_sampled
- contract_profile (incl. align_magnetic)
- rebin, rebin2d
- reflamp, magnetic_amplitude

The setup.py has been modified so it does not try to build binary extension.  It builds a wheel with extension py3-none-any (universal python3 wheel)

All the chisq tests return the same number as in the reflmodule.so branch, except for doc/examples/freemag/pmf.py
I haven't figured out why that one is slightly different.

refl1d/probe.py
```diff
@@ -1725,36 +1730,39 @@ def shared_beam(self, intensity=1, background=0,
                 x.sample_broadening = sample_broadening
 
     def oversample(self, n=6, seed=1):
+        self._oversample(n, seed)
+        self.oversampling = n
+        self.oversampling_seed = seed
```
pkienzle, line 1735: Is this going to conflict with the theta offset branch?

Maybe we should revert this file on this branch since it is limited to oversampling changes.

refl1d/lib_numba/magnetic.py
```diff
@@ -0,0 +1,397 @@
...
+        # minus to plus (lower to higher potential energy) and the observed R-+ will
+        # actually be zero at large distances from the interface.
+        #
+        # In the backing, the -S1 and -S3 waves are explicitly set to be zero amplitude
+        # by the boundary conditions (neutrons only incident in the fronting medium - no
+        # source of neutrons below).
+        #
+        S1L = -sqrt(complex(PI4*(RHO[L]+RHOM[L]) - E0, -PI4*(fabs(IRHO[L])+EPS)))
+        S3L = -sqrt(complex(PI4*(RHO[L]-RHOM[L]) -
+                    E0, -PI4*(fabs(IRHO[L])+EPS)))
```
pkienzle, line 134: Break after comma to keep expressions together.

Why does this code use `-PI4...` while the code in `lib/magnetic.cc` was changed in this PR to use `+PI4...`?

refl1d/reflmodule.py
```diff
@@ -0,0 +1,11 @@
...
+from .lib_numba.magnetic import magnetic_amplitude as _magnetic_amplitude
+from .lib_numba.magnetic import calculate_u1_u3 as _calculate_u1_u3
+from .lib_numba.convolve import convolve_gaussian
+from .lib_numba.convolve import convolve_uniform
+from .lib_numba.convolve_sampled import convolve_sampled
+from .lib_numba.contract_profile import align_magnetic as _align_magnetic
+from .lib_numba.contract_profile import contract_by_area as _contract_by_area
+from .lib_numba.contract_profile import contract_mag as _contract_mag
+from .lib_numba.rebin import rebin_counts
+from .lib_numba.rebin import rebin_counts_2D
```
pkienzle, line 11: I'm not sure we need a "lib_numba" subdirectory. Assuming we keep it, maybe we should to push more things into the calculation library in case other projects want to use them. Otherwise it feels a bit like an historical accident rather than deliberate design. Regardless, we should probably drop the C library (maybe keeping it around somewhere for a future cuda version) and move lib_numba to lib. Not required for this PR. Put a code cleanup ticket if you think it needs it.


Tests were failing because of a name conflict: reflmodule.py and reflmodule.*.so were competing because the .so files were still in the directory. This will cause problems as we switch between branches and go back in time to check differences in results. I suggest renaming `reflmodule.py` to `refllib.py` or `_reflectivity.py` and updating `reflectivity.py` and `profile.py` accordingly.

pkienzle (commented):
Please verify the sign on the complex part of the refractive index for the magnetic version, which at a glance looks inconsistent between the numba and the C code.

We should add the following to setup.py `python_requires='>=3.8',`, or make it 3.7 and add  3.7 to the CI tests.

Rename `reflmodule.py` to something that won't conflict with the old .so file for those of us running an "inplace" version.

Address the minor comments as you see fit and feel free to pull.

refl1d/probe.py
bmaranville: This does not conflict with the theta_offset branch.  It is the same change made in that branch, so they merge just fine.  Could probably revert it as you say with no issues.

refl1d/reflmodule.py
bmaranville: I moved to the lib_numba layout so that one could switch between the versions easily - they I broke that by naming the aggregator "reflmodule.py"  There are also a couple places where we (collectively) changed the signatures/interfaces of the functions and you can no longer easily switch back and forth.  I don't know how high a priority it is to make it possible to switch back.

The tests are only failing for you because you've previously built the C version - on a clean build (in the CI scripts) the tests are all passing.

refl1d/lib_numba/magnetic.py
bmaranville: This is an excellent question, and I wish I had written down what I was doing.

refl1d/reflmodule.py
pkienzle: Yes, on a clean build the tests pass. I'm trying to avoid confusion later when we switch back to an ancient version to do some sort of test or check then forget that we have to remove the .so file when moving back to the current version.

.github/workflows/test.yml
```diff
@@ -14,14 +14,8 @@ jobs:
     strategy:
       matrix:
         config:
-          - { os: ubuntu-latest, py: 3.8, doc: 1 }
-          - { os: windows-latest, py: 3.6, whl: 1 }
-          - { os: windows-latest, py: 3.7, whl: 1 }
+          - { os: ubuntu-latest, py: 3.8, doc: 1, whl: 1 }
```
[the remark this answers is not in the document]
bmaranville: I think I was using 3.8 as the target for the dataclasses branch, but I don't remember why.  I don't think it's important - 3.7 is fine.

refl1d/lib_numba/contract_profile.py
```diff
@@ -0,0 +1,265 @@
+import numba
+import math
+
+Z_EPS = 1e-6
+
+ALIGN_MAGNETIC_SIG = 'i4(f8[:], f8[:], f8[:], f8[:], f8[:], f8[:], f8[:], f8[:], f8[:,:])'
+
+
+#@numba.njit(ALIGN_MAGNETIC_SIG, parallel=False, cache=True)
+@numba.njit(cache=True)
```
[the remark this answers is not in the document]
bmaranville: for some of these there was explicit polymorphism encoded in the wrapper, and it seemed like a reasonable policy in numba to address that by not specifying a signature. I did some testing and was not able to see any performance difference between signature/no signature. For this function I suspect the usage of the variables is pretty unambiguous to numba.

refl1d/reflmodule.py
bmaranville: It seems that python resolves .so files before .py files?  If it were the other way around that would be more convenient in this case, though maybe not in general.

refl1d/lib_numba/magnetic.py
bmaranville: I think I was testing the sign conventions, and never reverted the .cc file because it was no longer being used. The negative signs in this expression are designed to explicitly choose the branch for the sqrt. There might be a better way to do this.

refl1d/reflectivity.py
```diff
@@ -213,19 +223,19 @@ def magnetic_amplitude(kz,
...
-    #print(np.vstack((depth, np.hstack((sigma, np.nan)), rho, irho, rhoM, thetaM)).T)
+    # kz = -kz
+    # depth, rho, irho, sigma, rhoM, thetaM = [v[::-1] for v in (depth, rho, irho, sigma, rhoM, thetaM)]
+    depth, rho, irho, sigma = [_dense(a, 'd')
+                                      for a in (depth, rho, irho, sigma)]
+    # np.set_printoptions(linewidth=1000)
+    # print(np.vstack((depth, np.hstack((sigma, np.nan)), rho, irho, rhoM, thetaM)).T)
 
-    sld_b, u1, u3 = calculate_u1_u3(H, rhoM, thetaM, Aguide)
+    sld_b, u1, u3 = calculate_u1_u3_py(H, rhoM, thetaM, Aguide)
```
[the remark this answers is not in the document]
bmaranville: This was an oversight - I switched to the pure python version before the numba version was ready, then never switched back.

doc/genmods.py
```diff
@@ -60,7 +60,6 @@
     ('probe', 'Instrument probe'),
     ('profile', 'Model profile'),
     ('reflectivity', 'Reflectivity'),
-    ('reflmodule', 'Low level reflectivity calculations'),
```
bmaranville, line 63: @pkienzle - do I need to re-enable this (with a different name) with some documentation for the low-level calculations?  I'm not quite sure how to do that.

refl1d/experiment.py
```diff
@@ -110,12 +110,15 @@ def update(self):
 
     def residuals(self):
         if 'residuals' not in self._cache:
+            # Trigger reflectivity calculation even if there is no data to
+            # compare against so that we can profile simulation code, and
+            # so that simulation smoke tests are run more thoroughly.
+            QR = self.reflectivity()
```
bmaranville, line 116: This seems like an important change, outside the refactor with numba... I hope this is in the other branches too

refl1d/reflectivity.py
bmaranville: The auto code formatter in VS code puts in the space... and if we are going to try to follow pep8 I think it's too much work to try to fight both battles.

refl1d/reflectivity.py
bmaranville: I reformatted the whole thing with VS code autoformat.

refl1d/reflmodule.py
bmaranville: moved to refllib.py

doc/genmods.py
bmaranville: I think I figured it out...

refl1d/reflectivity.py
pkienzle: My solution would be to avoid the vscode formatter since it has already demonstrated in other contexts that it doesn't know how to format python code.

bmaranville: If you can suggest an alternative formatter to use in vscode I would be happy to use it, but when formatting thousands of lines we can't really do it by hand.  I just checked and what I'm using now is "autopep8".  I also have "black" and "yapf".  
I think some of the earlier issues I had with it were user error (on my part).  If you can find errors in the many "pep8" commits that I just pushed I will adjust my settings.

bmaranville: My earlier point was that it is unsustainable to try to maintain a style that is not automatically respected by autoformatters (or automatically enforced by them).  If we are going to enforce style, we have to rely on the tools.

refl1d/experiment.py
pkienzle: Not super important. In the past I've forced the call directly within the model file, but this didn't work when profiling.

doc/genmods.py
```diff
@@ -59,6 +59,7 @@
     ('polymer', 'Polymer models'),
     ('probe', 'Instrument probe'),
     ('profile', 'Model profile'),
+    ('refllib', 'Low level reflectivity calculations'),
```
pkienzle, line 62: This will pick up the docstrings from the functions defined in refllib but no module level docstrings. This may be good enough for our purposes and we don't need to figure out how to document subpackages "properly". We might drop a comment here to that effect.

bmaranville: I had to add all the functions to an __all__ list in refllib.py in order for them to get picked up.  I agree we can figure out how to improve the docstrings in the future.  When you say "module-level docstrings" I'm not sure what you mean - I added a docstring to the top of refllib.py and it seems to have picked that up (it's pretty minimal).  If you have something in mind for a comment/todo to add to the file please feel free to add!

refl1d/reflectivity.py
pkienzle: The main tool I've used on other projects is lint, with periodic manual cleanup of lint errors. We had a discussion on the sasmodels repo about a tool to check whether a commit increased the lint count and flag those that do. Seems mostly doable, but we didn't act on it.

Regarding auto-formatters, PEP 8 was intended to be prescriptive not descriptive. Code "readability" trumps formatting consistency. When vscode breaks a line in the middle of an expression rather than breaking at a comma that hurts readability. If you would like to use a code formatter please look for one that knows how to format code.

pkienzle: I see settings lets you choose between black, autopep8 and yapf. I haven't tried any of them.

PR #135, opened 2021-07-30
cuda implementation of magnetic reflectivity

pkienzle:
Here's a 4x slower version of magnetic reflectivity using numba.cuda.

Run using:
```python
python run.py doc/examples/magrough/model.py --profile --steps=50 | less -S
```

Feel free to play and try to make it run faster.

### DIFF
--- a/refl1d/reflectivity.py
+++ b/refl1d/reflectivity.py
@@ -619,6 +619,8 @@ def magnetic_amplitude_driver(d, sigma, rho, irho, rhoM, u1, u3, KZ, R):
     R[:] = gR.copy_to_host(stream=stream)
 
 if USE_CUDA:
+    dev = cuda.select_device(0)
+    print(dev)
     stream = cuda.stream()
     CR4XA_SIG_F = CR4XA_SIG.replace('f8','f4').replace('c16','c8')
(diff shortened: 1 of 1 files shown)


pkienzle:
Note that this implementation is single precision with no stability correction. If it were fast enough then it might be worth trying to improve it. Basically, divide each matrix entry by the maximum on the diagonal and multiply the final result by that product. At least, that's how I was able to compute reflectivity from 10 km thick samples for the non-polarized case.

Modern gaming cards are 30x slower for double precision; maybe I'm accidentally promoting floats to double and killing performance that way.

pkienzle:
Part of the problem appears to be communication overhead with the card. Change q length from 400 to 4000 and execution time changes very little (the RTX 2080 card has 4300 processors). For 40000 q values execution time is 10x, which makes sense given the number of processors.

Changing the layer data to fill a single matrix might help with the communication overhead, but that's a more involved code modification. The new memory layout may help performance on the card, making it easier to place it into shared memory so that access patterns don't matter so much.

Making the reflectivity calculation "asynchronous" so that you can compose the next layer matrix while the current calculation is running on the card would also help.

In any case, convolution dominates when number of q increases, so that, too, needs to be moved onto the card.

PR #149, opened 2022-03-16
convert + to ++ and - to -- in load4

bmaranville:


refl1d/probe.py
```diff
@@ -1372,6 +1377,10 @@ def load4(filename, keysep=":", sep=None, comment="#", name=None,
         data_by_xs = {strip_quotes(entry[0]["polarization"])
                       : _data_as_probe(entry, probe_args, **data_args)
                       for entry in entries}
+        if AUTO_PROMOTE_HALPOL:
+            for halfpol in ("+", "-"):
+                if halfpol in data_by_xs:
+                    data_by_xs[halfpol*2] = data_by_xs.pop(halfpol)
```
pkienzle, line 1383: Maybe want to warn if half-polarization is detected? Probably not.

We could check if halfpol*2 is already defined before promoting so we don't overwrite an existing ++ or –– cross-section, but I think that this is unlikely enough that we can blow it off.

The `halfpol*2` is slightly magical. It may be cleaner to just write:
```python
if '+' in data_by_xs:
    data_by_xs['++'] = data_by_xs.pop('+')
if '-' in data_by_xs:
    data_by_xs['--'] = data_by_xs.pop('-')
```
Or in anticipation of some other halfpol syntax:
```python
_KEY_MAP = {'+': '++', '?+': '++', '+?': '++', '-': '--', '?-': '--', '-?': '--'}
for old, new in _KEY_MAP.items():
    if value := data_by_xs.pop(old, None):
        data_by_xs[new] = value
```

glass-ships:
Worth revisiting? 

> Implement half-polarized probe, that sums e.g. ++ and +- theory xs to compare to + summed data

possibly related?

bmaranville:
Agreed: this is a half-measure that can be abandoned in favor of the more rigorous solution being worked on.

PR #150, opened 2022-04-12
Orso loader

bmaranville:
This implementation puts the ORSO loader inline with load4 - it's possible that having a completely separate function maintained in e.g. ```orsodata.py```, next to ```ncnrdata.py``` would make more sense.  What do you think, @pkienzle ?


pkienzle:
I think load4 should be pulled out of probe.py as well.

glass-ships:
status of this? Do we want to keep working on this branch?

bmaranville:
This was fixed

PR #151, opened 2022-06-24
Remove unpolarized probe from SA plot

purnimab:
When Polarized and Unpolarized probes exist, Unpolarized probe is plotted in "SA" views as default view (log scale) such that the SA for polarized probes cannot be viewed. Removes this from the plot.

### DIFF
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -578,7 +578,7 @@ class Probe(object):
         Note: :meth:`oversample` will remove the extra Q calculation
         points introduced by :meth:`critical_edge`.
         """
-        
+
         rng = numpy.random.RandomState(seed=seed)
         T = rng.normal(self.T[:, None], self.dT[:, None], size=(len(self.dT), n-1))
         L = rng.normal(self.L[:, None], self.dL[:, None], size=(len(self.dL), n-1))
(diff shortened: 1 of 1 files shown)


bmaranville:
@purnimab  - this PR seems to be based on a branch other than 'master'.  Can you start a new PR with just the changes you've listed in this commit, but based on 'master'?

purnimab:
Ah, I'm sorry, could it be because I haven't pulled from the main repo for
a while so it's just out of date?

bmaranville:
Ah - I think I was wrong.  I saw a commit that looked like it came from a different branch, but github says your 'master' is just behind this 'master' by 6 commits, so the merge should work fine.

pkienzle:
@purnimab can you please add the fix I suggest in #153? It should still run `Build_Unstable / test_and_build` but `publish` and `updateUnstableText` should be skipped.

pkienzle:
Regarding this PR, does the SA plot still work when the datasets do not contain unpolarized?

Wouldn't it be better to suppress the unpolarized plots when SA is requested rather than suppressing the SA?

purnimab:
@pkienzle Since PolarizedNeutronProbe has it's own plot() function which does call plot_SA(), the SA plot still works for the polarized datasets. This should just suppress SA for an unpolarized probe.

purnimab:
Also, what is the file you want me to edit?

pkienzle (approved):

Issue #81, opened 2019-09-10
show dependent parameters in parameter table


pkienzle:
Defining a parameter such as `s = Parameter(name='scale', value=1)` and assigning it to a model such as `layer2.thickness = scale*layer1.thickness`, then the scale parameter does not show up in the Parameters tab and so you can not turn it on/off or select its range.

purnimab:
Seconding this feature request :)

bmaranville:
Is this a bumps issue then?

bmaranville:
I think this was fixed in https://github.com/bumps/bumps/commit/39a005945eba49a5946a28cfa33f923adc9db1ff - but only in the serializable branch that has yet to be merged.

bmaranville:
No - I could be wrong - this maybe is supposed to be covered by the Operator class, which should list all dependent parameters.  Will have to look into it.

PR #152, opened 2022-06-24
pass all dependent parameters from probe and model to parameters() call

bmaranville:
Proposed fix for #81

### DIFF
--- a/refl1d/model.py
+++ b/refl1d/model.py
@@ -91,10 +91,10 @@ class Layer(object): # Abstract base class
         """
 
     def layer_parameters(self):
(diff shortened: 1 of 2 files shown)


bmaranville:
I don't think I like this solution after all... probably better to add a handler in bumps.gui.parameter_view for classes that have a .parameters method (so that it automatically expands and includes dependent parameters)?

glass-ships:
Something we plan on returning to / implementing in this branch?

glass-ships:
Is the fix you mention implemented elsewhere? Should we revive this?

PR #157, opened 2023-02-24
Added functionality to fit 'sum' (mean) and difference

purnimab:
Added a probe which contains sum and difference cross-sections, an experiment which calculates residuals based on these rather than pp and mm.
Added new data views to plot difference rather than just spin asymmetry.
Save from gui implemented.
Also includes error-in-variables.


bmaranville:
This is very complete!  Thanks for all the work you put into it.

PR #158, opened 2023-03-06
MAINT: rename next --> i_next

andyfaff:
Replaces the use of `next` with `i_next` because `next` is a keyword that probably should be used as a variable name


bmaranville:
This seems uncontroversial, and the code change looks fine.  The tests pass.  I'm inclined to accept the PR.

PR #175, opened 2024-04-02
Fix calc_Q for shared theta_offset

bmaranville:
In models with a shared `theta_offset` (where the parameter is tied, for example for mm and pp) the values of Q at which the model is evaluated were not properly calculated (`PolarizedNeutronProbe.calc_Q`).

The first time calc_Q is determined, `measurement_union` is called, and then probe.T is set to the xs.T + shared_offset.  Later calls to `_calculate_union` add the current theta_offset to that T, which is incorrect (should be added to the original probe.T).

This PR stores the original T (as `probe._To`) and when `_calculate_union()` is called for a shared `theta_offset`, it will add the current offset to the original T.

Note that currently oversampling is not compatible with `_calculate_union`.  Need to fix that separately (oversampling is ignored when calculating new values of `calc_Q` when `theta_offset` is a fittable parameter)

### DIFF
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -1733,19 +1733,19 @@ def measurement_union(xs):
     TL = set()
     for x in xs:
         if x is not None:
-            TL |= set(zip(x.T+x.theta_offset.value, x.dT, x.L, x.dL, x.dQ))
-    T, dT, L, dL, dQ = zip(*[item for item in TL])
-    T, dT, L, dL, dQ = [np.asarray(v) for v in (T, dT, L, dL, dQ)]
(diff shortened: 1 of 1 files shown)


bmaranville:
Note that the result of this bug is that `calc_Q` for `PolarizedNeutronProbe` does not always line up perfectly with data `Q` values, so that some interpolation is done (data Q is always interpolated to calc_Q, but it's easier when they line up so the interpolation is trivial).  It should not affect chisq or any correctness, but it might if the data is not sampled densely enough (the minimum chisq for a fit is sensitive to choice of `calc_Q` for undersampled `calc_Q` - this is a known effect).

The "right" thing to do is to either take data densely enough to cover the convolution shape of dQ at each Q, or oversample the `calc_Q` values.

PR #177, opened 2024-06-17
Custom Plot API

hoogerheide:
This is a dummy custom plot API allowing responsive plots of quantities derived from the model parameters.

Basically the plot definition function is stored in a new attribute of `problem`, `problem.custom_plot`, and must output a Plotly figure that is rendered directly and responsively in the Refl1D GUI.

Future improvements could involve allowing multiple plots (say, as a dictionary, with keys corresponding to the title of each custom plot, and they can be selected using pills or tabs).

This PR is intended mainly as a conversation starter around this. A model file with data that show this in operation is attached.

[crau_popc.zip](https://github.com/user-attachments/files/15870914/crau_popc.zip)

### DIFF
--- a/refl1d/webview/client/src/components/CustomPlot.vue
+++ b/refl1d/webview/client/src/components/CustomPlot.vue
@@ -1,0 +1,51 @@
+<script setup lang="ts">
+/// <reference types="@types/plotly.js" />
+import { ref, shallowRef } from 'vue';
+import * as Plotly from 'plotly.js/lib/core';
+import type { AsyncSocket } from 'bumps-webview-client/src/asyncSocket.ts';
+import { setupDrawLoop } from 'bumps-webview-client/src/setupDrawLoop';
+import { COLORS } from '../colors.mjs';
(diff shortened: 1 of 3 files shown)


bmaranville:
This is an interesting proposal, and adding custom plots to the interface would definitely be useful.  I can see how attaching them directly to a problem would make sense too, since the type of plot may be very specific or even unique.

I'm not quite sure how to make this portable, i.e. how do you share the custom plot routines unless you require that the model always be loaded from your python script?  You can pickle the functions, but that is somewhat fragile, and then you lose the capability of representing your problem in terms of a plain dictionary (the basis for building a model builder).

I'm guessing that the plots you are building are not arbitrary, and are in fact quite specific and relevant to the models built with molgroups, in which case we should definitely support Experiment or FitProblem subclasses providing their own plotting routines, so that all users can make use of these.

hoogerheide:
Superceded by https://github.com/reflectometry/refl1d/pull/199

PR #181, opened 2024-09-08
Webview

pkienzle:
update to the webview branch


pkienzle:
Tests are passing locally. Need to do some fiddling with github actions so that it doesn't try running old refl1d against new bumps.

pkienzle:
Still need a code review of all the changes

refl1d/errors.py
```diff
@@ -357,34 +385,36 @@ def _write_file(path, data, title, columns):
...
-            _draw_overplot(group, 3, name + ' rhoM')
+            _draw_overplot(group, 3, name + ' rhoM', axes=axes)
         else:
-            next_color()
-    _profile_labels()
+            next_color(axes=axes)
+    _profile_labels(axes=axes)
 
-def _draw_overplot(group, index, label):
+def _draw_overplot(group, index, label, axes=None):
```
pkienzle, line 405: Passing fig and axes around everywhere (here and in bumps) is tedious.

refl1d/experiment.py
```diff
@@ -379,11 +388,19 @@ def __init__(self, sample=None, probe=None, name=None,
...
         self._slabs = profile.Microslabs(num_slabs, dz=dz)
         self._probe_cache = material.ProbeCache(probe)
         self._cache = {}  # Cache calculated profiles/reflectivities
-        self._name = name
+        self.name = name if name is not None else probe.name
+        self.constraints = constraints
+        self.version = __version__ if version is None else version
+        if auto_tag:
+            tag_all(self.probe.parameters(), 'probe')
+            tag_all(self.sample.parameters(), 'sample')
```
pkienzle, line 403: For simplicity I collapse the parameters into a flat list when I walk the model in bumps. Maybe this tagging could happen during flattening?

refl1d/errors.py
bmaranville: Yes!  and also thread-safe.

refl1d/experiment.py
bmaranville: What would be gained by doing the tagging during flattening?  You would have to keep track of the "auto_tag" setting (it would become part of the state of the class).

pkienzle: If every application using bumps serialization needs tagging, then tagging is a bumps concern, not an application concern. [Not that I know what tagging does, but it doesn't seem like it would be specific to reflectometry.]

pkienzle:
I think I can fix the errors by replacing `master` with `master_v0` in the actions. This should clean up the errors so that we can merge this back to main. Still need to fix the refl1d and bumps webview builders to target master rather than webview.

Feel free to rename the branch from `master_v0` to `master_wx` or `wx_master`.

PR #185, opened 2024-09-12
Restructure and Reorganize project

glass-ships:
Restructures refl1d to house files together in folders related to their function/purpose

To Do:
- [x] Fix docs not building (since `MODULES` list assumes all files are top level apparently)
- [x] Double check docs are appropriately updated


glass-ships:
This is currently failing for an interesting reason. 
Installing the package doesn't seem to include the newly created folders, and I'm not entirely sure why. 

```sh
$ pwd
.../envs/refl1d/lib/python3.11/site-packages/refl1d 

$  ls
fitplugin.py  __init__.py  lib  main.py  names.py  __pycache__  view  webview
```

Maybe @bmaranville or @pkienzle has some insight?

bmaranville:
It looks like package discovery is disabled if we manually provide a list of packages in the `pyproject.toml` file, so we can either add all the new packages (folders) to that list, or remove it completely and let auto-discovery happen.

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#automatic-discovery

bmaranville:
I feel like the primary use of the refl1d library is to construct an Experiment object, each of which requires a Sample object and a Probe object, so having experiment.py and probe.py at the same level inside refl1d/models and sample-building stuff in a subfolder refl1d/models/sample/ doesn't seem to match the hierarchy implied in the class construction.  Why not experiment.py in refl1d/ and then tools for constructing samples in refl1d/sample/ and probes in refl1d/probe/ ?

```mermaid
flowchart BT
    Sample --> Experiment 
    Probe --> Experiment
```

glass-ships:
> I feel like the primary use of the refl1d library is to construct an Experiment object, each of which requires a Sample object and a Probe object, so having experiment.py and probe.py at the same level inside refl1d/models and sample-building stuff in a subfolder refl1d/models/sample/ doesn't seem to match the hierarchy implied in the class construction. Why not experiment.py in refl1d/ and then tools for constructing samples in refl1d/sample/ and probes in refl1d/probe/ ?

so that's similar to what i had in mind with this PR. the idea behind what i have so far is that experiment, probe, and sample, are all models of "things" that make up a particular instance refl1d project, rather than a literal representation of what's made up of what else, and that sample is a directory is that there are multiple "subtypes" of sample, whereas experiment and probe (along with instrument and fitproblem) are represented by single files, so there's not really anything else that would go into `probe/` other than `probe.py` which would feel unnecessarily stuttery. 

experiment _could_ live at the top level, it just made sense to classify it as a model in the context of "python model representing a 'thing'". but we can definitely move some things around as it makes sense to

bmaranville:
Also, `backends.py` doesn't really go in `fitting`... I'm trying to think where that belongs.  There was some importlib reason that it's not in the lib folder directly, but the calculations that go in the backends aren't really specific to the act of fitting, they are just accelerated version of the calculations of the reflectivity (which is done in Experiment), the profile (which is calculated based on the sample), etc. 

The backend chooser is just there to select which version of the fast calculations is used in the various places around the library that need to go fast.

glass-ships:
> Also, `backends.py` doesn't really go in `fitting`... I'm trying to think where that belongs. There was some importlib reason that it's not in the lib folder directly, but the calculations that go in the backends aren't really specific to the act of fitting, they are just accelerated version of the calculations of the reflectivity (which is done in Experiment), the profile (which is calculated based on the sample), etc.
> 
> The backend chooser is just there to select which version of the fast calculations is used in the various places around the library that need to go fast.

ah interesting, ok. my assumption was that it was choosing which backend was ultimately used to run the calculations as part of the fitting process. in that case, it might even make sense to just live at the root directory

Makefile
```diff
@@ -13,3 +13,12 @@ lint: ## Run ruff linting
...
 	@ruff format refl1d/ tests/ setup.py
+
+.PHONY: clean
+clean: ## Delete some cruft from builds/testing/etc.
+	rm -f `find . -type f -name '*.py[co]'`
+	rm -rf `find . -name __pycache__ -o -name "*.egg-info"` \
+		`find . -name 'output-*'` \
+		.coverage build dist doc/_build \
+		.pytest_cache \
+		.ruff_cache
```
pkienzle, line 24: Will this work on windows? [Does it matter?]

Are we sure we cannot have spaces in any of the found names?

glass-ships: referencing line 21 specifically? 
this is just to clean up the files written out by running `pytest` locally:
```
output-expt.json
output-profile.dat
output-refl.dat
output-slabs.dat
output-steps.dat
```

I'm not sure that this would work on windows, which does matter for developers on windows. i'm not familiar with cmd / powershell or how exactly to do os checks in a Makefile, but if you have ideas we can expand on this.

bmaranville: Why is there a Makefile in the root folder of the project now?  It does not build the project - can it live somewhere else?  Having documentation (this file exists to do <...things>) in a comment at the top of the Makefile would be helpful, too.

glass-ships: root is a pretty common home for Makefiles, especially ones that contain convenience functions, in this case cleaning up files (many of which are in or relative to the root directory), as well as formatting and linting. 

having it here also makes it extensible if we do find we want to add convenience targets for building, installing, running in dev mode, compiling the web client, etc etc. 

we could add documentation at the top, but typically i just define a `help` target which prints out the targets contained and what they do, [for example](https://github.com/monarch-initiative/monarch-app/blob/main/Makefile#L8) (although I've recently grown fond of the much slimmer one-liner which grabs comments with double-# and uses those)

bmaranville: For the actions that require significant path manipulations, we could make a separate `bash` script and call that from the Makefile.  

`bash` is available on Windows with `conda install -c conda-forge git-bash` or by installing git system-wide, and importantly `bash` is available on Windows Github action runners by default.  We could make it a requirement for a dev environment on Windows?

glass-ships: We could yeah. I do tend to prefer root level, as its pretty natural, common, and transparent/easy to find, read, extend, etc. But we totally can if you think it makes sense for this project.

glass-ships:
I just noticed there's an additional requirements.txt in `doc/`, is this necessary?

bmaranville:
> I just noticed there's an additional requirements.txt in `doc/`, is this necessary?

That file is directly referenced by `.readthedocs.yaml` in the root folder, and is part of the template structure at https://github.com/readthedocs/tutorial-template/

bmaranville:
I think you're right, it is redundant.  Probably the `.readthedocs.yaml` could be updated so that it uses the requirements from `pyproject.toml`, but I'm not entirely sure how to do that

doc/getting_started/install.rst
```diff
@@ -25,91 +28,160 @@ The installed python is a full version of python. If your specialized
...
+    pip install .[webview]
 
-With most pypi packages now bundled with wheels, it is now easy to set up a
-development environment using the official python package.  Similarly,
-anaconda provides binaries for all the refl1d dependencies.
+Additional Configuration - Windows 
+``````````````````````````````````
 
-You will need a C/C++ compiler.  If you already have Microsoft Visual C
+To install Refl1D on Windows, you will need a C/C++ compiler.  If you already have Microsoft Visual C
```
bmaranville, line 96: Since we switched to numba as the default backend, no additional compiler (beyond the one bundled in numba) is needed for refl1d on any platform.  Any references to compilers in this document can be removed.

doc/getting_started/install.rst
```diff
@@ -25,91 +28,160 @@ The installed python is a full version of python. If your specialized
...
 
-Linux
------
+Additional Configuration - OS/X
+```````````````````````````````
 
-Linux distributions will provide the base required packages.  You
-will need to refer to your distribution documentation for details.
+Similar to windows, you can install the official python distribution or
+use Anaconda.  You will need to install the Xcode command line utilities
```
bmaranville, line 110: as mentioned above, xcode is not needed.

andyfaff: Miniforge is probably better, given that anaconda/miniconda require licences for most institutions.

bmaranville: Agreed!  miniforge is great.

andyfaff: https://discuss.scientific-python.org/t/response-to-anaconda-switch-to-paid-plans/1395/11

glass-ships:
> I think you're right, it is redundant. Probably the `.readthedocs.yaml` could be updated so that it uses the requirements from `pyproject.toml`, but I'm not entirely sure how to do that

After some research it seems like the RTD team is averse to ~getting with the times~ supporting additional / modern file types, so pyproject.toml isn't an option. 
But depending on how the Jenkins job is set up, it may not be important. 

ie. provided that it does a `pip install refl1d[dev]` or `.[dev]`, it'll install sphinx and should be able to build the docs from there. if we want/need to pin sphinx to a particular version, we can do that in the pyproject.toml

Makefile
glass-ships: what's the verdict on this btw - ok to leave it top level, or prefer to move into a subdir?

bmaranville:
My understanding is that readthedocs does its own docs build - and it needs a configuration file to install the required dependencies before it does so.  I'm not sure how to tell it to install the [dev] dependencies.

I'd kinda like to build our own docs - it's a pretty easy thing to do (there are GH actions for sphinx builds, and I've used them on other projects)

glass-ships:
> My understanding is that readthedocs does its own docs build - and it needs a configuration file to install the required dependencies before it does so.  I'm not sure how to tell it to install the [dev] dependencies.
> 
> I'd kinda like to build our own docs - it's a pretty easy thing to do (there are GH actions for sphinx builds, and I've used them on other projects)

that's definitely an option as well, maybe worth investing a little time into setting that up? probably in a separate pr

bmaranville:
> > My understanding is that readthedocs does its own docs build - and it needs a configuration file to install the required dependencies before it does so.  I'm not sure how to tell it to install the [dev] dependencies.
> > I'd kinda like to build our own docs - it's a pretty easy thing to do (there are GH actions for sphinx builds, and I've used them on other projects)
> 
> that's definitely an option as well, maybe worth investing a little time into setting that up? probably in a separate pr

See #194

glass-ships:
phew ok, I think this should be ready for a final review! 
finally finished combing through the install and contribution docs under getting_started

bmaranville:
I am working on a very detailed review right now - I will post it here shortly (next week?)

refl1d/readers/load4.py
bmaranville, line 1: This is fantastic - exactly the right thing to do, to pull this function out.  Might also want to split out a separate ORSO loader, but can do that later.

bmaranville:
Ok, here is my first attempt at a summary of the package, which hopefully explains the directory structure proposed at the end (which is based on the reorg proposed in this PR)

# Refl1D Package
The package, installable from pypi, includes both a library for defining reflectometry models and applications that use that library.

## applications
### CLI (command-line interface)
A longstanding interface for running fits and simulations from the command line, inheriting most behaviors from the bumps command-line-interface.  Access is through `main.py`

### garefl.py
A shim for running models defined for the C-based `garefl` fitting program in Refl1D

### WX Gui
A WXPython-based graphical interface for interactive fitting and visualization.  Not being actively developed, and possibly deprecated.

### webview
A web-based graphical interface for interactive fitting, implementing all the features found in the previous WXPython GUI.


## defining your models
### sample model
_representing the physical thin-film object under study_

- create multiple layers, including:
   - scattering power (SLD)
   - absorption
   - roughness
   - thickness
   - (optionally) magnetism
- sometimes, define a layer according to a function
- sometimes, define a scattering power from a material description
- finally, combine layers into a stack

### probe model
_representing the reflectometry measurement_
- indicate the angles and wavelengths (or Q) of the (real or simulated) measurement
- describe the resolution of measurement point
- describe the magnetic field (magnitude and direction)
- optionally, include the actual data measured at each point (using one of the loaders)

### experiment model
_bundle the measurement definition with the corresponding proposed sample definition_ 
  - include at least one sample model
  - include at least one probe model
  - indicate some optimization values to be used during calculations

## evaluating the models
This follows after the models are defined, of course

### sample model evaluation
1. render the layers into a set of "profiles"
   - profiles are one-dimensional maps of the following quantities vs. depth into the sample (z, units of distance)
      - scattering power (SLD)
      - absorption
      - (optionally) magnetic moment and direction
   - the roughness defined between layers determines the blending of those quantities between layers
   - input parameters: (typically supplied by Experiment model)
      - activating microslabbing (boolean) `step_interfaces`
      - setting z-step for microslabbed profiles `dz`
   - setting re-aggregation parameter `dA` for combining similar microslabs after render (performance optimization, reduces number of calculations of reflectivity)
1. Optionally, recombine similar microslabs into larger layers according to similarity criterion parameter `dA` (supplied by Experiment model)

### probe model evaluation
_at the moment, all these calculations are pretty simple and are done inline, not using accelerated versions in `refl1d.lib`_
- calculate a set of unique Q values covering all cross-sections in a polarized probe (only for polarized probes) `Q`
- oversample Q values to provide sufficient basis for smearing calculated R by resolution function `calc_Q`

### calculating resolution-corrected reflectivity
1. Take outputs of
   - sample model evaluation (profiles) and
   - probe model evaluation (set of Q values, `calc_Q`)
   
   and use them to calculate R at each requested value of Q, using the `lib.reflectivity` or`lib.magnetic` functions from the chosen backend
2. Apply resolution smearing to R vs. `calc_Q` to get R vs. `Q`

### calculating residuals
(only if data is attached to the probe objects)
- subtract calculated resolution-corrected `R` from measured `R` at each `Q`, and divide by `dR` to give residuals

## What bumps does to run a fit with these models:
- modifies parameters marked as fittable, e.g. layer thicknesses, scattering power, roughness...
- evaluates the models, which triggers 
   - rendering new profiles
   - possibly altering the `calc_Q` values
   - calculating reflectivity
   - resolution correction
   - subtraction from data to get residuals

The sum of the residuals is divided by the number of degrees of freedom in the model (the number of free parameters + number of data points) to give the reduced chi-squared parameter, which is a measure of the goodness of fit that can be used in Bayesian probability analysis.

## Other uses of the models
- interactive manual adjustment of parameters "chi by eye"
   - often done prior to a bumps fit
- experiment simulation (experiment optimization)

## Proposed new layout:

```
./
├── applications/
│   ├── garefl.py
│   ├── main.py
│   ├── webview/
│   │   ├── build_client.py
│   │   ├── client/
│   │   └── server/
│   └── wx_gui/
│       └── view/
├── models/
│   ├── experiment.py
│   ├── names.py
│   ├── uncertainty.py
│   ├── bumps_interface/
│   │   ├── fitplugin.py
│   │   └── fitproblem.py
│   ├── lib/
│   │   ├── refllib.py
│   │   ├── c_ext/
│   │   ├── numba/
│   │   └── python/
│   ├── probe/
│   │   ├── fresnel.py
│   │   ├── instrument.py
│   │   ├── oversampling.py
│   │   ├── probe.py
│   │   ├── resolution.py
│   │   ├── stitch.py
│   │   └── data_loaders/
│   │       ├── anstodata.py
│   │       ├── load4.py
│   │       ├── ncnrdata.py
│   │       ├── snsdata.py
│   │       ├── staj.py
│   │       └── stajconvert.py
│   └── sample/
│       ├── cheby.py
│       ├── fasta.py
│       ├── flayer.py
│       ├── layers.py
│       ├── magnetism.py
│       ├── material.py
│       ├── materialdb.py
│       ├── mono.py
│       └── polymer.py
└── tests/
    ├── abeles.py
    ├── gepore.f
    ├── gepore_zeeman.f
    └── refl_tr.py
```

### Notes:
- I think it might be useful to deprecate `refl1d/models/names.py` and put all those names in `__init__.py`, so that users could do `from refl1d.models import *` instead.
- I'm not sure if `tests` should go under `models` or at the top level - probably need a tests folder at the top level since we test the applications and the library.
- the `refl1d/view` folder is moved to `refl1d/applications/wx_gui/view`

bmaranville:
Also I don't really know if the `applications` folder is really necessary, or if we just want to put all those at the root folder level.

glass-ships:
- I agree tests should go top level, that's pretty typical especially since the tests aren't localized to the models and span the majority of the repository.
- having `view` under `wx_gui` seems somewhat redundant - I would propose simply moving the contents of `view` into `wx_gui` directly
- i'm not sure i entirely agree with `lib/` going under `models/`, as it's not really functional classes representing "objects", they're more "backend" functionality that make sense to live in their own house

outside of those suggestions, i think i mostly agree with your proposed structure

bmaranville:
> * I agree tests should go top level, that's pretty typical especially since the tests aren't localized to the models and span the majority of the repository.
> 
>     * having `view` under `wx_gui` seems somewhat redundant - I would propose simply moving the contents of `view` into `wx_gui` directly
> 
>     * i'm not sure i entirely agree with `lib/` going under `models/`, as it's not really functional classes representing "objects", they're more "backend" functionality that make sense to live in their own house
> 
> 
> outside of those suggestions, i think i mostly agree with your proposed structure

The mapping to folders is not purely based on the semantics of building models, but also evaluating them and transforming them, etc.  It would be overly cluttered to make a separate parallel tree of modules for "evaluate_models", I think.

Since the modules in `lib` exist only to perform the calculations used within the models, I thought they might get bundled in that folder for practical reasons.

glass-ships:
personally i don't think it would be overly cluttered, at the moment there are only three directories under the top level `refl1d/` - applications, models, and tests - and their functionality could (read: should) be documented for clarity. 

but i defer to your judgement there! 

i'll look to take some time in the next couple weeks to restructure the repo again

glass-ships:
another quick question - I thought we checked and confirmed that `garefl.py` is no longer needed?

glass-ships:
also, where would rebin.py go?

bmaranville:
`rebin.py` is only used by the data_loaders modules, so probably putting it in that folder makes sense.  Though there is an accelerated version in `lib` that we could use, and remove the standalone `rebin.py` module.  It would take a little refactoring of the code in the loaders that use it.

Let me double-check with our main users of garefl here - to see if they still want to keep it in.  Some of them were not at the code camp.

bmaranville:
As for separating things further between model building and evaluation:

In simple cases, methods are defined inline in the model class for doing calculations, and as they get more complicated, they have been split out into separate modules.  There is therefore a fuzzy boundary between them, and they can't always be cleanly separated, and I'm not sure there's much benefit to doing so.

glass-ships:
Second restructure complete, tests updated and fixed. 
only remaining issues:
- wx won't install
- need to update documentation

glass-ships:
come to think of it... was pytest running against the wx stuff before?

glass-ships:
yeah I'm guessing it's a pytest configuration thing, I forgot to check the project.toml. 
will do first thing in the morning

glass-ships:
could be shortened to 
```python
from refl1d.probe.probe import Probe`

probe = Probe(...)
```
but yeahhh it's a bit silly

glass-ships:
building docs fails due to warning: 
```
.../refl1d/refl1d/models/bumps_interface/fitproblem.py:docstring of refl1d.models.bumps_interface.fitproblem.FitProblem:10: WARNING: undefined label: 'freevariables' [ref.ref]
```

which is true, because refl1d's fitproblem doesn't have that in the docstrings, while bumps's fitproblem does.. not sure how to fix.

glass-ships:
other than that, this is ready for re-re-review

glass-ships:
Commented out replacing `FitProblem` docstring for bumps's version until that can be resolved separately. 

Tests working, updated with main - should be ready for review

glass-ships:
@bmaranville slightly unrelated. 
`webview/client/package.json` lists `"vue-json-viewer": "^3.0.4"`
but this npm package lists `2.22.2` as latest, and i don't see it imported anywhere in the client code... can it safely be removed?

bmaranville:
I think the piece that was using it was moved to the bumps codebase... and then we switched to a different library to do the same thing... so yes, removable.

glass-ships:
@bmaranville looking at [doc/guide/index.rst L94](https://github.com/reflectometry/refl1d/blob/restructure/doc/guide/index.rst?plain=1#L94):
the `PTFit` class doesn't seem to exist anymore (seems to have been moved to bumps?) - how should this be modified?

bmaranville:
> @bmaranville looking at [doc/guide/index.rst L94](https://github.com/reflectometry/refl1d/blob/restructure/doc/guide/index.rst?plain=1#L94): the `PTFit` class doesn't seem to exist anymore (seems to have been moved to bumps?) - how should this be modified?
> 
> Same thing with the `_MaterialStacker` class in `models.sample.material.Scatterer`

I would just replace the PTFit with something like the amoeba fitter class, but I don't really know what `_MaterialStacker` was.  That reference might be very old.

glass-ships:
> I would just replace the PTFit with something like the amoeba fitter class

this is also not in refl1d, would that be `bumps.fitters.SimplexFit`?

bmaranville:
I started running existing models with the restructured code, and the biggest observation I made is that the majority of them use functions that are not in `refl1d.names`.  Most of the time, this is something related to implementing a custom data loader, such as for TOF data, that uses `make_probe`.  Other times utility functions like `bumps.util.make_seed` are being used, or the `FWHM2sigma` function from `refl1d.resolution`.

Some of these are unnecessary - `make_probe` is a minuscule function that just returns one of the Probe subclasses, and they could be used directly (and are in the `names` namespace), and we could easily add something like `FHWM2sigma` to `names`.

So, the takeaway is that the majority of existing scripts would break with this change, and fixing them would be fairly easy for one of the developers, but a big pain in the neck for users.

The main reason I haven't pushed harder to get this merged is that the new folder structure still doesn't completely map onto the graph of how the modules work and interact - it is hard to sort them in this way, and there are many interconnections, and I don't think any folder structure can accurately represent this.  The practical grouping of e.g. "this module has to do with probe stuff (building, calculating, using...)" is the best we can do.  Do you feel like this meets the initial goals of the PR?  Is it worth the cost to the users, of having to modify old scripts in order to have them work, which many of them will not know how to do?

glass-ships:
I do feel strongly that some sort of sensible grouping of modules is important, it's pretty common for things in one folder to interact with things in another folder and that doesn't really break any sense of logic imo, but it does make sense to, like you said, group things together at the very least in _some_ way, ie (these modules relate to the probe, these modules relate to the sample, etc etc).  
If there are any other changes to the folder structure you'd suggest, I'm all ears, even removing the `models` dir and moving everything underneath it to the top level as previously mentioned is sensible.

Breakage could be mitigated by adding whatever we need to to `names` to maintain consistent functionality.  
A migration guide could also go a long way to helping, as well as informative deprecation warnings and error messages.

Personally, I'm ok with things breaking now and then if it's a side effect of improving the code base and making it more maintainable/easier to understand - these things do happen from time to time, what's more important is providing a helpful path forward for users/developers. 

But again, that's just one developers opinion

glass-ships:
I do feel strongly that some sort of sensible grouping of modules is important, it's pretty common for things in one folder to interact with things in another folder and that doesn't really break any sense of logic imo, but it does make sense to, like you said, group things together at the very least in _some_ way (ie. these modules relate to the probe, these modules relate to the sample, etc etc).  

If there are any other changes to the folder structure you'd suggest, I'm all ears, even removing the `models` dir and moving everything underneath it to the top level as previously mentioned is sensible.

Breakage could be mitigated by adding whatever we need to to `names` to maintain consistent functionality. A migration guide could also go a long way to helping, as well as informative deprecation warnings and error messages.

Personally, I'm ok with things breaking now and then if it's a side effect of improving the code base and making it more maintainable/easier to understand - these things do happen from time to time, what's more important is providing a helpful path forward for users/developers. 

But again, that's just one developers opinion

PR #188, opened 2024-09-13
Get version number from tag using versioningit

backmari:
This change eliminates the need for a hard coded version number, since the package version number will be generated from the git tag.

Same change done recently in bumps: https://github.com/bumps/bumps/pull/153


backmari:
@bmaranville can you please review?

bmaranville (approved):

PR #192, opened 2024-09-17
update Zenodo badge to generic version

bmaranville:
According to the settings page on Zenodo, this badge image and DOI link will always resolve to the latest respective version.


bmaranville:
This is pretty uncontroversial and it would be nice to have the correct badge showing... merging without review.

PR #196, opened 2024-09-25
remove approximation shortcut for magnetic_amplitude

bmaranville:
(when rhoM < MINIMAL_RHO_M)

Users noticed an incorrect calculation of the spin asymmetry when the applied field was less than the equivalent MINIMAL_RHO_M (when H < 0.00425 T).  The reason for the incorrectness is a shortcut that we were taking to speed up calculations when rhoM in the fronting or backing medium is small, but 

- this shortcut only speeds up the calculations by a factor of 2
- it seems to be noticeably wrong whenever it is applied, getting worse the higher the field is
- reducing MINIMAL_RHO_M would make it less and less wrong, but would also make the shortcut only apply in rarer and rarer cases, to the point that it's not worth including it at all.

so in the interest of correctness, we should probably remove the shortcut

### DIFF
--- a/refl1d/lib/c/magnetic.cc
+++ b/refl1d/lib/c/magnetic.cc
@@ -12,7 +12,6 @@
 #define M_PI 3.141592653589793
 #endif
 
-#define MINIMAL_RHO_M 1e-2  // in units of 1e-6/A^2
 const double EPS = std::numeric_limits<double>::epsilon();
 const double B2SLD = 2.31604654;  // Scattering factor for B field 1e-6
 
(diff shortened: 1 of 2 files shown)


bmaranville:
The other option would be to manually clamp rhoM to be zero in the fronting and backing medium if it is below some cutoff, but since it has been demonstrated that having an incorrect value of rhoM leads to noticeable problems (having the wrong initial energy, because of the wrong rhoM, for the minus cross sections is why the shortcut results are wrong) it may not be worth this optimization, but it's something to keep in mind if it becomes really important to speed up calculations of reflectivities in very small fields (and the errors from clamping rhoM in the fronting will be different, i.e. they won't cause a spurious spin-asymmetry)

bmaranville:
It's true that in the case where M is purely parallel or antiparallel to H, the Zeeman energy term is just an arbitrary offset to the total energy that cancels because the spin state is the same after reflection as it was in the incident beam... And in that case it would be reasonable to do the shortcut after setting the Zeeman energy to zero.

So if we do decide to keep this shortcut, the right guard condition is probably to check if any of the magnetization in any layer deviates from the H quantization axis (to guard against spin flip)

acaruana2009 (approved):

bmaranville:
It would probably be even faster to fall back to the strategy of just calculating two separate reflectivities for `++` and `--` cross-sections using the "unpolarized" kernel with `rho + rhoM` and `rho - rhoM` when we assert that spin-flip can't happen, so this shortcut can be removed for good I think.

PR #197, opened 2024-09-26
Experiment plot hooks for custom plotting

hoogerheide:
Creates the appropriate fields in Experiment class to be able to use new bumps custom plotting infrastructure.

PR #199, opened 2024-10-01
Custom plots rebase

bmaranville:
To replace #197

### DIFF
--- a/refl1d/experiment.py
+++ b/refl1d/experiment.py
@@ -340,8 +340,10 @@ class ExperimentBase:
     def register_webview_plot(
         self, plot_title: str, plot_function: WebviewPlotFunction, change_with: Literal["parameter", "uncertainty"]
     ):
-        # Plot function syntax: f(model, problem, state)
         # change_with = 'parameter' or 'uncertainty'
+        # Plot function syntax: f(model, problem) for 'parameter'
+        #                       f(model, problem, state, n_samples) for 'uncertainty'
+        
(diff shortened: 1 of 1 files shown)


refl1d/experiment.py
```diff
@@ -323,6 +337,17 @@ def save_refl(self, basename):
             theory = self.reflectivity(interpolation=self.interpolation)
             self.probe.save(filename=basename + "-refl-interp.dat", theory=theory)
 
+    def register_webview_plot(
+        self, plot_title: str, plot_function: WebviewPlotFunction, change_with: Literal["parameter", "uncertainty"]
+    ):
+        # Plot function syntax: f(model, problem, state)
```
hoogerheide, line 343: Now the plot function syntax depends on the function: for "parameter" functions it's just `f(model, problem)` and for "uncertainty" functions it's `f(model, problem, state, n_samples)`

This comment should probably either be removed or modified.

hoogerheide (changes requested):
Nice generalization of the plotting functions that keeps everything separate from the webview GUI it is intended to serve.

See the one inline comment about documentation, which is now out of date.

refl1d/experiment.py
hoogerheide: Resolved by https://github.com/reflectometry/refl1d/pull/199/commits/0743a4a794c0fedff8509ddcb73cebeb5a0442c6

hoogerheide (approved):

PR #203, opened 2024-10-08
Update release actions

bmaranville:
Getting actions ready for publishing a wheel to pypi, and for building distributables for all platforms.  Tested on a fork, all of the actions seem to be working.


bmaranville:
pre-commit.ci run

bmaranville:
I'm realizing that the publish-pypi action won't work for pre-releases that well... it is looking for the latest release of bumps to install, while we want it to use the minimum required bumps version from pyproject.toml (which sometimes will be a pre-release)

I'm working on a fix now, using the builder script in `refl1d/webview/build_client.py` instead of specifying the steps for building in the action directly, because if we pip install from source, we'll get the correct version of bumps, and build_client will use that.

bmaranville:
I think I will merge this without review, because it's impossible to test its full functionality without merging it and attempting to publish a new pre-release to pypi.

PR #208, opened 2024-10-14
Windows shell launcher

bmaranville:
A bat file so that users can open a terminal window with the refl1d virtual environment activated, to enable simple installation of additional refl1d-related package, e.g.
```sh
pip install molgroups
```


bmaranville:
This is a pretty minimal change, and it would be nice to test (it can't really break anything)

PR #210, opened 2024-10-16
Remove client cdn

bmaranville:
This PR implements a matching change to the one in `bumps` in this PR: https://github.com/bumps/bumps/pull/172

The fallback CDN for `refl1d-webview-client` that had previously been published on npmjs, and mirrored with cdn.jsdelivr.net, would be removed from the webserver code.

It would then be required that client code exist at the path `refl1d/webview/client/dist/`. 

- Users installing from pip (pypi) will have this code included in the distributed package automatically
- Developers running from an editable install will have to run `python -m refl1d.webview.build_client` every time the client source changes
   - they will also need to pull a new version of bumps in their editable install, prior to running the build_client command.  It is not necessary to build the bumps client to build the refl1d client, just make sure the appropriate source files are in place in the bumps install.

Merging this PR will have to be synchronized with merging https://github.com/bumps/bumps/pull/172


bmaranville:
Ok, this is a mess after I rebased it locally and pushed... I will clean up...

glass-ships (approved):

PR #211, opened 2024-10-22
add flags for enabling optional persistent path from shortcuts

bmaranville:
Using the persistent path written to disk is now optional, but we want to enable that feature for users running the webview application from a shortcut that was added during installation of the conda-packed distribution.


bmaranville:
requires bumps >= 1.0.0a6 but that requirement has already been changed in pyproject.toml

hoogerheide (approved):
Looks good. It makes sense for persistence to be on in canned environments and off in the general case.

PR #213, opened 2024-10-24
Make LICENSE.txt consistent with documentation

bmaranville:
The license (3-clause BSD) contained in the documentation under `doc/getting_started/license.rst` was not the same as the content of `LICENSE.txt`

This PR adds the 3-clause BSD license to LICENSE.txt, and includes that file directly into the `license.rst` with an `include` statement.

Also, the `CREDITS` section is updated to include the new collaboration, and the citation is updated to use the DOI created by Zenodo.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,1 +1,21 @@
-Refl1d is in the public domain.
+Copyright 2006-2024 The Refl1D Contributors
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the 
+following conditions are met:
+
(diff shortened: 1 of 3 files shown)


bmaranville:
I tried to update the classifier for the license to a more specific one, but I think it will fail to upload to pypi if it doesn't match one of the ones here:
https://pypi.org/classifiers/ so I reverted to "License :: OSI Approved :: BSD License"

mdoucet (approved):
This looks very good!

PR #218, opened 2024-11-05
allow background parameter of Probe to be negative

bmaranville:
When fitting data that has been "oversubtracted", sometimes the apparent background level is negative.

Setting a hard lower "physical" limit of 0.0 for the background parameter makes it hard to fit this type of dataset, without manually adjusting the limits after the Probe object is created

Currently, you would have to do e.g. this, to fit a negative apparent background (after initializing `Probe` object):
```python
probe.background.limits = (-float("inf"), float("inf"))
```

### DIFF
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -758,7 +758,7 @@ class Probe(BaseProbe):
             name = os.path.splitext(os.path.basename(filename))[0]
         qualifier = " " + name if name is not None else ""
         self.intensity = Parameter.default(intensity, name="intensity" + qualifier)
-        self.background = Parameter.default(background, name="background" + qualifier, limits=(0.0, None))
+        self.background = Parameter.default(background, name="background" + qualifier, limits=(None, None))
         self.back_absorption = Parameter.default(back_absorption, name="back_absorption" + qualifier, limits=(0.0, 1.0))
         self.theta_offset = Parameter.default(theta_offset, name="theta_offset" + qualifier)
         self.sample_broadening = Parameter.default(sample_broadening, name="sample_broadening" + qualifier)
--- a/refl1d/probe.py
+++ b/refl1d/probe.py
@@ -758,7 +758,7 @@ class Probe(BaseProbe):
             name = os.path.splitext(os.path.basename(filename))[0]
         qualifier = " " + name if name is not None else ""
         self.intensity = Parameter.default(intensity, name="intensity" + qualifier)
-        self.background = Parameter.default(background, name="background" + qualifier, limits=(None, None))
+        self.background = Parameter.default(background, name="background" + qualifier)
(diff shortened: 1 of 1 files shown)


acaruana2009:
I think this is fine, I have discussed it with some instrument scientists in my group and they are fine with it as a pragmatic solution to the problem.

acaruana2009 (approved):

alexander-grutter:
One issue I've been wrestling with is how to display the data. Obviously negative reflectivity is nonsense and also it's hard to put on a log scale. What do we think about shifting the data by the fitted background offset? Feels iffy but all other solutions feel worse.

alexander-grutter:
That's actually a feature I've wanted before. Again, it feels iffy, but I can see the value if applied judiciously.

hoogerheide:
I have used a symmetric log plot for this in the past: https://journals.iucr.org/j/issues/2022/01/00/ge5108/

I really like these plots, mainly because you can see all of the negative data and, importantly, whether there is any structure in the negative data. The proposal to just shift the background will still hide all of the negative data, though I do think it's a good start.

The downside is that they are somewhat difficult to generate. Matplotlib has symlog scaling but I don't know how to do this in plotly.

bmaranville:
Symlog has been requested from plotly for the past 8+ years... no movement to implement as far as I can tell.

acaruana2009:
> I have used a symmetric log plot for this in the past: https://journals.iucr.org/j/issues/2022/01/00/ge5108/
> 
> I really like these plots, mainly because you can see all of the negative data and, importantly, whether there is any structure in the negative data. The proposal to just shift the background will still hide all of the negative data, though I do think it's a good start.
> 
> The downside is that they are somewhat difficult to generate. Matplotlib has symlog scaling but I don't know how to do this in plotly.

That is super nice! For spin flip data this would be very useful indeed! I have never even thought about plotting our data up in a symmetric log plot.

bmaranville:
Ok, I looked at plotly in depth and adding symlog to that is a major project, not something we can do.

Should we merge this and make a new issue for how to modify the display, or does everyone agree on what should be shown for the reflectivity wrt. background?

acaruana2009:
I am happy for this to be merged.

andyfaff:
A comment from the sidelines. I frequently model XRR data on a logR transformed scale, because I find that the uncertainties that are often assigned during reduction aren't as good as they could be. A logR transform seems to work quite well. In that case a negative background doesn't make sense, and won't work with a logR transform. Oversubtraction of background is probably not an issue for XRR? 

One thing I've thought about is that if you know your background has to be positive, then it's numerically more stable in a fit to have the background specified in log-space, i.e. `limits=[-8, -5]`, rather than in lin-space, `limits=[10**-8, 10**-5]`.
I wonder if it's still possible to have the magnitude of the background specified in log-space if the background can be negative.

PR #219, opened 2024-11-06
make a serializable version of FunctionalProfile and FunctionalMagnetism

bmaranville:
These versions of FunctionalProfile and FunctionalMagnetism differ from the original:

- if the classes are initialized with a `profile` function, they will read the source with `inspect.getsource`.
- if the classes are initialized with `profile_source` string, the string will be executed in a context with `scipy` and `np` defined, and the function that results will be pulled out.  _(the function is expected to be the only defined name in the locals context when the exec is complete, so the string should contain only a function definition and nothing else)_
- profile functions can only use basic math operators and anything in the `np` and `scipy` namespaces (these are embedded in the execution context), e.g.
```python
def blobby(z, a, b):
    return scipy.special.erf(b) * np.sin(a * z)
```
- parameters for the function are not added to the class as attributes
- they are instead added to a dict attribute `FunctionalProfile.profile_params`
- new attributes are added which are Parameter objects, and a Calculation object in their `slot` attribute:
   - `rho_start`
   - `irho_start`
   - `rho_end`
   - `irho_end`
- the `start` and `end` attributes are now properties, which create an SLD object, e.g. `start == SLD(rho=rho_start, irho=irho_start)`

For `FunctionalMagnetism`, a further difference is that the total thickness is stored as an expression in the class.  You can set it with a `set_anchor` function as before, but now that function will construct a thickness expression, or you can pass in an existing Parameter or Expression, e.g. `layer[3].thickness + layer[4].thickness`

### DIFF
--- a/doc/examples/profile/model.py
+++ b/doc/examples/profile/model.py
@@ -100,14 +100,14 @@ flayer.magnetism.set_anchor(sample, "sin")
 # Set the fittable parameters.  Note that the parameters to the function
 # after the first parameter *z* become fittable parameters.
 
-sample["sin"].period.range(0, 100)
-sample["sin"].phase.range(0, 1)
 sample["sin"].thickness.range(0, 1000)
-sample["sin"].magnetism.M1.range(0, 10)
-sample["sin"].magnetism.M2.range(0, 10)
-sample["sin"].magnetism.M3.range(0, 10)
-sample["sin"].magnetism.z1.range(0, 100)
-sample["sin"].magnetism.z2.range(0, 100)
+sample["sin"].profile_params["period"].range(0, 100)
+sample["sin"].profile_params["phase"].range(0, 1)
+sample["sin"].magnetism.profile_params["M1"].range(0, 10)
+sample["sin"].magnetism.profile_params["M2"].range(0, 10)
+sample["sin"].magnetism.profile_params["M3"].range(0, 10)
+sample["sin"].magnetism.profile_params["z1"].range(0, 100)
+sample["sin"].magnetism.profile_params["z2"].range(0, 100)
 
 # Define the model.  Since this is a simulation, we need to define the
 # incident beam in terms of angles, wavelengths and dispersion.  This
--- a/refl1d/flayer.py
+++ b/refl1d/flayer.py
@@ -126,10 +126,12 @@ class FunctionalProfile(Layer):
         self.tol = tol
         self.magnetism = magnetism
 
-        if profile_params is not None:
-            self.profile_params = profile_params
+        if profile_params is None:
+            # use kw arguments to set initial values for profile parameters
+            self.profile_params = _init_profile_params(name, self.profile, kw, self.RESERVED)
         else:
-            _set_parameters(self, name, self.profile, kw, self.RESERVED)
+            # no need to filter out reserved keywords
+            self.profile_params = _init_profile_params(name, self.profile, profile_params, ())
 
         # TODO: maybe make these lazy (and for magnetism below as well)
         self._set_ends()
@@ -171,7 +173,7 @@ class FunctionalProfile(Layer):
                 "thickness": self.thickness,
                 "interface": self.interface,
                 "profile": self.profile,
-                "parameters": _get_parameters(self),
+                "parameters": self.profile_params,
                 "tol": self.tol,
                 "magnetism": self.magnetism,
             }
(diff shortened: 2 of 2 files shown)


bmaranville:
Given that these changes will probably break the way that e.g. molgroups use `FunctionalProfile`, we should probably offer them as a new type rather than replacing the existing classes.

acaruana2009:
> Given that these changes will probably break the way that e.g. molgroups use `FunctionalProfile`, we should probably offer them as a new type rather than replacing the existing classes.

I am happy with this. How would it work in principle? Would we just have a different name for the class/type?

bmaranville:
Actually, the way dataclass equality comparison works the result is 
```python
(layer.start == layer.start) is True
```
(it compares attributes, and they're all equal)

The dynamic attributes don't specifically interfere with serialization, but add maintenance overhead for the developers (weird bugs!) and cognitive overhead for the users (not a common pattern in python anymore, and full of footguns). 

The rho_start etc. are by default `None`, and I agree most users won't initialize them directly.

pkienzle:
This doesn't capture the variables from the script used inside the function.

Here's some code to run in a jupyter cell to explore the space:
```python
import numpy as np
from dataclasses import dataclass
from typing import Any
@dataclass
class CapturedFunction:
    name: str
    source: str
    context: dict[str, Any]
    def __init__(self, name, source, context):
        self.name = name
        self.source = source
        self.context = context
        self._fn = None

    @property
    def fn(self):
        if self._fn is None:
            import numba
            global_context = {
                'np': np,
                'njit': numba.njit,
                'prange': numba.prange,
                **self.context,
            }
            local_context = {}
            #print(f"compiling <{self.source}> in", self.context)
            exec(self.source, global_context, local_context)
            #print("local context", local_context)
            self._fn = local_context[self.name]
        return self._fn

    def __call__(self, *args, **kw):
        return self.fn(*args, **kw)
    
def capture_source(fn):
    if isinstance(fn, CapturedFunction):
        return fn
    
    import inspect
    from textwrap import dedent
    #from copy import deepcopy
    #import bumps.pmath
    #whitelist = bumps.pmath.__all__
    base_context = {'np': np} #{k: getattr(np, k) for k in whitelist}

    name = fn.__name__
    print("type fn", type(fn))
    source = dedent(inspect.getsource(fn))
    print("source =>", source)
    context = {}
    try:
        closure = inspect.getclosurevars(fn)
        print("closure =>", closure)
        # build context from closure
        # ignoring builtins and unbound for now
        for k, v in closure.globals.items():
            if k not in base_context:
                context[k] = v # TODO: deepcopy(v) ?
        for k, v in closure.nonlocals.items():
            if k not in base_context:
                context[k] = v # TODO: deepcopy(v) ?
    except Exception as exc:
        print("exception capturing closure", exc)
    
    capture = CapturedFunction(name, source, context)
    capture._fn = fn # already have the function; don't need to recompile
    return capture

cutoff = 5
def topf(x, a=1, b=5):
    return np.minimum(a*x + b, cutoff)
def do():

    def roundtrip(fn, kwargs):
        import inspect
        value = fn(**kwargs)

        #print("1")
        capture = capture_source(fn)
        print("source => ", capture.source)
        print("context => ", capture.context)
        #assert (value == capture.fn(**kwargs)).all()

        # Force recompile
        capture._fn = None
        assert (value == capture.fn(**kwargs)).all()

    kwargs = dict(a=15, b=3, x=np.arange(5))

    print("== simple ==")
    def f(x, a=1, b=5): return a*x + b
    roundtrip(f, kwargs)
    
    # Can't do lambda because the function object doesn't have a name
    # Because inspect.getsource is so stupid, we would have to extract the
    # lambda expression from a source line such as f = lambda x, a, b: ...
    # or roundtrip(lambda x, a, b: ..., kwargs). Can't do either of these
    # without a full parser.
    if 0:
        print("== lambda ==")
        f = lambda x, a=1, b=5: a*x + b
        roundtrip(f, kwargs)
        print("== lambda ==")
        roundtrip(
            lambda x, a=1, b=5: \
            a*x + b,
            kwargs,
        )
    
    print("== function ==")
    def f(x, a=1, b=5): return np.sin(a*x + b)
    roundtrip(f, kwargs)
        
    print("== closure over locals ==")
    cutoff = 36
    def f(x, a=1, b=5): return np.minimum(a*x + b, cutoff)
    roundtrip(f, kwargs)

    print("== closure over globals ==")
    roundtrip(topf, kwargs)

    print("== numba jit with prange ==")
    from numba import njit, prange
    # Numba doesn't work with closures. It uses the value
    # as defined when njit was called. I don't see anything
    # obvious in the jitted object that contains the context.
    #cutoff = 4
    @njit(parallel=True)
    def f(x, a=1, b=5):
        cutoff = 4
        result = np.empty_like(x)
        for k in prange(len(x)):
            if x[k] < cutoff:
                result[k] = a*x[k] + b
            else:
                result[k] = cutoff
        return result
    #print(f(**kwargs))
    #print("jit", dir(f))
    #print(f.py_func.__code__)
    roundtrip(f, kwargs)

    # helper functions
    # class method
    # instance method
    # helper class
    # third party packages like pandas

do()
```

bmaranville:
closing... not much interest from the community, and #276 overlaps with intent of this

PR #220, opened 2024-11-07
Configure linting for frontend

glass-ships:
- [x] Adds eslint and prettier configs
- [x] Adds a `lint` and `test:lint` script in `package.json`
    - Can be run with `bun|npm run test:lint` to preview warnings/errors, or `bun|npm run lint` to apply fixable changes
- [x] Applies formatting changes
- [x] Address unfixable linting errors
- [ ] Add make target for formatting/linting
- [ ] Automate linting to pre-commit or a github action

EDIT: Will likely open a separate PR that isolates these changes and merges into main, rather than restructure. Any additional changes needed should go into #185 or its own PR after the fact.


refl1d/webview/client/src/standalone_signal.ts
```diff
@@ -1,42 +1,42 @@
 export class Signal {
-    name: string;
```
bmaranville, line 2: It's worth double-checking, but I don't think this code is used anywhere else in this repo anymore, and we might be able to remove it.

glass-ships: i believe you are correct!

bmaranville: It got moved over here: https://github.com/reflectometry/refl1d-demo

.pre-commit-config.yaml
```diff
@@ -17,4 +17,4 @@ repos:
       - id: ruff-format
 
 ci:
-    autoupdate_schedule: monthly
\ No newline at end of file
+  autoupdate_schedule: monthly
```
bmaranville, line 20: would this setting not have worked with the existing indent?

glass-ships: oh it would, i didn't notice it changed... that's not from the eslint/prettier stuff, that was just me doing my habitual formatting hotkey spam. 
we can totally revert it

refl1d/webview/client/bun.lockb
bmaranville, line 1: I am not sure this was supposed to be committed.

PR #227, opened 2024-11-14
Formatting changes, update imports, other minor changes

glass-ships:


glass-ships:
Unclear why a few doc tests are failing.

glass-ships:
This somehow introduces an issue where `plot_data` is empty so plots don't get drawn. 
Investigating.

glass-ships:
@mdoucet or @bmaranville I could use a little help debugging this if you get a chance

glass-ships:
Resolved the above issue - forgot that the unused imports in `webserver.py` aren't actually entirely unused.

doc/getting_started/contributing.rst
```diff
@@ -55,13 +88,13 @@ As you make changes to the package, you can see what you have done using git::
...
-More thorough tests are found in test directory.  With the nosetest package,
-you can run the tests using::
-
-    python tests.py
+doctests on all of our examples so that we know our documentation is correct.
+More thorough tests are found in test directory. You can run these tests via pytest, 
+or via the convenience Makefile target::
 
-Nose is available on linux form apt-get
+    pytest
```
bmaranville, line 95: Does this have to be `pytest -v`?

doc/getting_started/contributing.rst
```diff
@@ -113,7 +143,7 @@ Building Documentation
 ======================
 
 Building the package documentation requires a working Sphinx installation,
-and latex to build the pdf. As of this writing we are using sphinx 1.2.
+and latex to build the pdf. As of this writing we are using sphinx 8.0.2.
```
bmaranville, line 146: wow that's a big jump...

doc/getting_started/contributing.rst
```diff
@@ -125,7 +155,7 @@ You can see the result by pointing your browser to::
     doc/_build/latex/Refl1d.pdf
 
 Note that this only works with a unix-like environment for now since we are
-using *make*.  On windows, you can run sphinx directly from python::
+using *make*.  On Windows, you can run sphinx directly from python::
```
bmaranville, line 158: I think make is pretty easy to install from conda-forge now... might simplify things on Windows, though I haven't tested it for building the documentation.  Might be additional dependencies to install as well.

doc/getting_started/install.rst
```diff
@@ -25,91 +28,136 @@ The installed python is a full version of python. If your specialized
...
-Linux distributions will provide the base required packages.  You
-will need to refer to your distribution documentation for details.
+    cd /path/to/refl1d/refl1d/webview/client
+    npm install                     # or bun install
+    npm link bumps-webview-client   # or bun link bumps-webview-client
 
-On debian/ubuntu, the command will be something like::
+To build the client, run::
 
-    sudo apt-get install python3-{matplotlib,numpy,scipy,wxgtk4.0,pyparsing,setuptools}
```
bmaranville, line 98: Maybe list `python -m refl1d.webview.build_client` here?

doc/getting_started/install.rst
```diff
@@ -25,91 +28,136 @@ The installed python is a full version of python. If your specialized
...
+This starts the client and shows the URL to connect to in the terminal (typically http://localhost:5173).
 
-OS/X
-----
+Now, you can start the Python webview server with::
 
-Similar to windows, you can install the official python distribution or
-use Anaconda.  You will need to install the Xcode command line utilities
-to get the compiler.
+    refl1d-webview --headless --port 8080
```
bmaranville, line 159: I would not recommend `--headless` as the default... if they leave that off, it will open a browser for them automatically, using the correct localhost url.

doc/pylit.py
bmaranville, line 1: I honestly don't know what this file is for.

doc/getting_started/contributing.rst
glass-ships: it does not

explore/interface.py
```diff
@@ -93,8 +93,12 @@ def default(cls, value, **kw):
                 return cls(value)
 
 
-sech = lambda x: 1 / cosh(x)
-asech = lambda x: acosh(1 / x)
+def sech(x):
```
bmaranville, line 96: this change results in a different object for `sech` and `asech` - does it affect their intended use?

doc/getting_started/contributing.rst
glass-ships: yeah, i didn't actually update anything either, documentation just hadn't been updated since then

doc/getting_started/install.rst
```diff
@@ -25,91 +28,136 @@ The installed python is a full version of python. If your specialized
...
+    
+    # if you also want to run the webview, an optional extra is available
+    pip install refl1d[webview]
 
-Note that the binary versions will lag the release version until the release
-process is automated.  Windows and Mac users may want to install using pip as
-well to get the version with the latest
-`changes <https://github.com/reflectometry/refl1d/blob/master/CHANGES.rst>`_.
+.. note ::
+    The binary versions will lag the release version until the release
```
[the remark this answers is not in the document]
glass-ships: sweet - note removed!

refl1d/polymer.py
```diff
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-# This program is public domain
-# Authors Paul Kienzle, Richard Sheridan
```
bmaranville, line 3: If effort was made to add attribution, I'd leave it in, as it doesn't hurt anything.  If we want to move this specific callout it to the license page or the documentation page that talks about attribution, that would be fine

refl1d/probe.py
```diff
@@ -1,7 +1,4 @@
-# coding=utf-8
-# This program is in the public domain
```
bmaranville, line 2: This attribution and others like it can be removed... we have already provided attribution for Paul in the docs.

refl1d/probe.py
```diff
@@ -12,7 +9,9 @@
 
 """
 
-from __future__ import with_statement, division, print_function
+from __future__ import division, print_function, with_statement
```
bmaranville, line 12: I don't think the `__future__` import is needed anymore

doc/getting_started/install.rst
glass-ships: My understanding was that going to `localhost:8080` took you to the version that uses the pre-compiled JS, rather than the local Vue/TS which is getting served at `localhost:5173/?server=localhost:8080`

This section is meant to be "if you are also developing the javascript environment. 

I can add an additional reference without `--headless` in the section above

refl1d/view/data_view.py
```diff
@@ -1,25 +1,19 @@
...
-
+from bumps.fitproblem import FitProblem
+from bumps.gui.util import EmbeddedPylab
 from matplotlib.backends.backend_wxagg import FigureCanvasWxAgg as FigureCanvas
 from matplotlib.backends.backend_wxagg import NavigationToolbar2WxAgg as Toolbar
 
 # The Figure object is used to create backend-independent plot representations.
 from matplotlib.figure import Figure
-
-import matplotlib.pyplot as plt
```
bmaranville, line 16: This is probably fine, and we aren't really supporting wx gui anymore, but the order of import statements might matter, if the backend is being specified in one of the imports.

bmaranville (changes requested):
Looks really good - a few minor changes suggested inline.

refl1d/view/data_view.py
glass-ships: Unfortunately I was never able to get wx installed because of lack of python version support, so I can't really test this. Can revert it if you prefer

explore/interface.py
bmaranville: Ah, ok thanks.  In some places in bumps we do very subtle things where the difference between a lambda and a function matters a lot, but not here.

doc/pylit.py
glass-ships: the only other place pylit gets used is in `gentut.py`, but I dont think it references this file. safe to delete?

refl1d/view/data_view.py
bmaranville: I think Paul approved complete deprecation of the WX gui in version 1.0, so it's up to you.  Not much point in fixing the formatting of the code, but not much harm either.

Removal of the `view` folder will take a bit of work - there are pieces that get used in `export results`, for instance, that have to be excised first.

doc/pylit.py
bmaranville: I don't completely understand how the documentation gets created, tbh.  Is there any chance sphinx makes stuff from here?

refl1d/view/data_view.py
glass-ships: cool ok, may as well leave it then, and work on just removing `view/` later on. i don't expect any further changes to files here formatting wise

doc/pylit.py
glass-ships: I believe sphinx references `conf.py` which calls `gentut.py` to generate the tutorials. but `gentut.py` just imports [pylit](https://pypi.org/project/pylit/), which is just a direct copy of an old version of https://codeberg.org/milde/pylit/src/branch/master/pylit.py

We can either leave pylit.py as it is, or delete it and add `pylit` to the pyproject.toml as a dev dependency.

I tested the latter option, and it seems to work just fine, so either option would be fine. Dealer's choice

PR #228, opened 2024-11-15
Add frontend linting

glass-ships:
New version of #220 based on `master` rather than `restructure`.

- [x] Adds eslint and prettier configs
- [x] Adds a `lint` and `test:lint` script in `package.json`
    - Can be run with `bun|npm run test:lint` to preview warnings/errors, or `bun|npm run lint` to apply fixable changes
- [x] Adds a `format` script in `package.json` to only run prettier without linting
- [x] Applies formatting changes
- [x] Address unfixable linting errors
- [x] Add make target for formatting/linting
    - Address bun vs npm
- [x] Automate linting to pre-commit or a github action
    - Same as above


bmaranville:
How would this work for applying formatting changes?  Is there a pre-commit hook that will do formatting like ruff does for python?

glass-ships:
unfortunately there isn't currently a pre-commit hook for running eslint/prettier via npm or anything, so the idea is that the `test-webview-client` workflow would run `bun run test:lint` and if there are errors/formatting changes to apply, the action will fail, so the user would have to run `bun run lint` or `bun run format` locally to address those, then push the changes. 

it is somewhat annoyingly manual, but the other option would be adding another formatting workflow that runs `bun run format` and pushes the changes back to the same branch, which could potentially get a little messy. i'm ok with either option, really

bmaranville:
I see the commented-out local runner for eslint in `.pre-commit-config.yaml` 

Is that something that could be made to work?

glass-ships:
possible, but it previously failed because npm/bun aren't installed on the pre-commit CI runner. could try adding a line that installs it

bmaranville (approved):

PR #231, opened 2024-11-20
Test types

glass-ships:
Adds type testing to the webview client via `vue-tsc`


glass-ships:
Still probably want to add type testing to Bumps and fix things there first, but. 
For whatever reason, refl1d isn't loading against the `lint-frontend` branch of bumps.  
Console log shows:  
`Uncaught (in promise) DOMException: String contains an invalid character`

Worth investigating as we'll probably need to fix it eventually. 
Unclear where the issue originates, but the bottom of the trace stack claims it starts at `main.js:29`. 

Additionally, getting a warning: 
```
[Vue warn]: Unhandled error during execution of component update 
  at <App panels= 
Array(15) [ {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, … ]
 name="Refl1D" >
```
but again, no further info about what that unhandled error might be.

bmaranville:
can you build the client with --sourcemap and see if the console indicates which Promise is failing?

glass-ships:
builds just fine:
```
$ vite build --sourcemap --emptyOutDir -m development
vite v6.0.2 building for development...
✓ 1451 modules transformed.
dist/0.1.26/index.html                                     0.61 kB │ gzip:   0.38 kB
dist/0.1.26/assets/refl1d-icon_48x48x32.HJicgF6C.png       4.07 kB
dist/0.1.26/assets/bumps-icon_256x256x32.Dz4RP7uH.png     54.62 kB
dist/0.1.26/assets/index.BcuciB-q.css                    359.00 kB │ gzip:  48.67 kB
dist/0.1.26/assets/index.BGn7BLFK.js                   2,751.94 kB │ gzip: 946.38 kB │ map: 11,359.74 kB

(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
✓ built in 8.40s
```

but running `vite preview` the page still doesn't load, and console doesn't print anything other than:
```
GET http://localhost:4173/ [HTTP/1.1 404 Not Found 1ms]
GET http://localhost:4173/favicon.ico [HTTP/1.1 404 Not Found 0ms]
```

bmaranville:
404 is probably an error on the python side... not serving a file being asked for.  I'll check it out.

bmaranville:
It's getting hung up on the `gear.svg` content (there is some illegal string in there that is causing the client error).  If I remove `<Gear />` from the template for App.vue in bumps, the refl1d client builds and loads fine.

What's really weird is that the bumps client builds and loads just fine with the gear.svg external resource... it's only when importing App.vue into refl1d that the string is getting mangled.

bmaranville:
ah... it looks like the svgLoader plugin is missing in the vite.config.js for refl1d

glass-ships:
wow, how the heck did you find that out?
good catch, I wonder why it's getting caught up, since it works in bumps... maybe some issue with vite svg loader or something

glass-ships:
ohhhh yeah, that would do it

bmaranville:
I did some debugging in the browser console... trapped the exceptions and looked at the variables when the exception was thrown, and saw that the string was an svg definition, and small, looking like an icon, then looked for places we are using inline svg.

glass-ships:
nice yeah, you'll have to show me how you did the trapping bit, that would have had me stuck for days

glass-ships:
Types test now passes against bumps's `test-types` branch

refl1d/webview/client/package.json
```diff
@@ -9,12 +9,13 @@
   },
   "scripts": {
     "dev": "vite",
-    "build": "vite build --emptyOutDir -m development",
+    "build": "vite build --sourcemap --emptyOutDir -m development",
```
bmaranville, line 12: I don't think we want to build sourcemaps all the time - it will slow things down and bloat the distributed packages. You can always do `npm run build -- --sourcemap` locally to test things out, and there's a flag to the python helper for doing that too.

bmaranville (approved):
These look good

PR #232, opened 2024-11-21
Merge FreeMagnetismInterface feature enhancement

acaruana2009:
`FreeMagnetismInterface` acts in the same way that `FreeInterface` does. It allows for a freeform connecting spline between magnetic layers to deal with non-trivial magnetic interfaces. 

Additional modifications have been put into `mono.py` in order to make it more useable, however, more work is probably needed in all of the spline functions we have in refl1d - this may be out of scope of this PR.


refl1d/magnetism.py
```diff
@@ -437,3 +437,150 @@ def __str__(self):
...
+    the above and below layers respectively. Do not specify if
+    layers either side are not magnetic.
+
+    *tbelow* and *tabove* are the thetaM parameter values of
+    the above and below layers respectively. Do not specify if
+    layers either side are not magnetic.
+    """
+
+    name: Optional[str]
+    mbelow: Optional[Union[Parameter, float]]
```
acaruana2009, line 478: @bmaranville I am not sure if this is the current style that you are using for typing. I.e. would you prefer to remove `Any` and `Optional`?

bmaranville: It's not really a question of style - the dataclass attribute types should be exactly what will be stored in those attributes after init + post_init.  There can be different types for the arguments of the init function, if they are going to be coerced before being stored in the attribute.  For example, an argument will often be of type `Union[float, Parameter]` but then we store it as `Parameter.default(value)` which coerces it to a Parameter, so that should be the type of the attribute (not the Union)

bmaranville: The typing of the arg is not the same as the typing of the attribute - `tbelow` and `tabove` are both initialized as `Parameter` objects during init, even if they are `None` as an argument.  The type of the attributes should be `Parameter`

bmaranville:
@acaruana2009 have you tried the new class out?  Does it work as you expect?  Can you serialize/deserialize it?

acaruana2009:
> @acaruana2009 have you tried the new class out? Does it work as you expect? Can you serialize/deserialize it?

Yes, it all seems to work. I haven't done any fits with it yet, but I can load in a file, save out into a session and reload into a session and it keeps chisq intact.

refl1d/mono.py
```diff
@@ -115,16 +117,19 @@ class FreeInterface(Layer):
...
+    name: str
+    below: Scatterer
+    above: Scatterer
     thickness: Par
     interface: Par
-    dz: List[Union[float, Par]]
-    dp: List[Union[float, Par]]
+    dz: List[Par]
+    dp: List[Par]
+    magnetism: Optional[BaseMagnetism] = None
```
bmaranville, line 127: Can this really be any `BaseMagnetism` object, or does it have to be specifically a `FreeMagnetismInterface`?

bmaranville (approved):

PR #234, opened 2024-11-21
show y axis of reflectivity plots in scientific notation (10 to a power)

bmaranville:
It has been requested by @hoogerheide that the y-axis number formatting for reflectivity plots be changed to a power of 10, e.g. $10^{-6}$

# What it looks like now:
image

# What it would look like after this PR:
image

### DIFF
--- a/refl1d/webview/client/src/components/DataView.vue
+++ b/refl1d/webview/client/src/components/DataView.vue
@@ -366,7 +366,7 @@ async function draw_plot() {
         x: 0.8,
         yanchor: "top",
         y: -0.05,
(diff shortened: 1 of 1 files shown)


mdoucet (approved):
Looks good. Should we change the chisq to $\chi^2$?

bmaranville:
Oh that's a good question... Not sure if plotly supports unicode annotations

hoogerheide:
It definitely does...I use one for Angstroms in the molgroups custom plots

acaruana2009 (approved):
Looks good!

PR #238, opened 2024-12-09
fasta is part of periodictable and isn't needed in refl1d

pkienzle:


bmaranville:
Well that certainly solves the problem.

bmaranville (approved):

PR #241, opened 2024-12-12
Add Refl1D migration between schemas

bmaranville:
- Adds a specific migration function to migrate models written before the restructure change to the new class layout.  
- Also adds the infrastructure to support further migrations.

In the future, if a breaking change is made which will affect the serialized output, the version number in `refl1d.__schema_version__` can be incremented, and a new function added to the lookup table that will migrate the old JSON structure to one compatible with the new.

### DIFF
--- a/refl1d/bumps_interface/migrations.py
+++ b/refl1d/bumps_interface/migrations.py
@@ -1,0 +1,75 @@
+from copy import deepcopy
+from typing import Optional, Tuple
+import re
+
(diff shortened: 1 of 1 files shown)


bmaranville:
Note: won't actually do anything unless PR https://github.com/bumps/bumps/pull/204 is merged first.

glass-ships (approved):

PR #244, opened 2025-01-03
Pytest plain asserts

bmaranville:
There is a conflict between the most recent scipy version (1.15) and the assertion-rewrite feature of pytest at the moment.  Until this is resolved, we can use plain assertions (no rewrite) with pytest.

The only impact is that failed assertion messages might be slightly less informative.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "bumps>=1.0.0a11",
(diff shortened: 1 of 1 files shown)


glass-ships (approved):

bmaranville:
I think based on the analysis of the ticket that was done in the scipy repo, that we can set up our pytest to be more targeted (it seems we're testing the entire codebase of scipy and all our other dependencies at the moment) - and if we do that we can probably turn assert rewriting back on.

PR #250, opened 2025-01-22
add serializable attributes to capture critical edge oversampling

bmaranville:


hoogerheide (approved):
This looks very good. It also looks like an expert user can add a custom `OversampledRegion` which is a nice feature.

bmaranville:
Yes, I was hoping this would be extensible!

pkienzle:
I just added #262 requesting a different point generator for uniform rather than normal ΔQ distribution.

In practice the wings of one sample set done using the normal distribution will show up in the resolution window for the neighbouring point so this is not a big issue. Only at certain places, such as at the edges of a gap in the data, will this lead to wasted theory calculations and poorly supported ΔQ.

refl1d/probe/probe.py
```diff
@@ -1417,19 +1440,32 @@ def scattering_factors(self, material, density):
...
     oversample.__doc__ = Probe.oversample.__doc__
 
     def critical_edge(self, substrate=None, surface=None, n=51, delta=0.25):
         Q_c = self.Q_c(substrate, surface)
-        extra = np.linspace(Q_c * (1 - delta), Q_c * (1 + delta), n)
-        calc_Q = np.hstack((self.Q, extra, 0))
-        self.calc_Qo = np.sort(calc_Q)
+        region = OversampledRegion(Q_start=Q_c * (1 - delta), Q_end=Q_c * (1 + delta), n=n)
+        self.oversampled_regions.append(region)
+        self._apply_oversamplings()
```
pkienzle, line 1472: Calling twice, for example in a jupyter notebook when you are exploring the choice of delta, will create a doubly oversampled part. I suggest keeping the critical region oversampling separate from the additional oversampled regions.

pkienzle: Throw an error if there is already an oversampling specified, and perhaps warn that the method is deprecated.

PR #253, opened 2025-01-28
Aguide convention fix

bmaranville:
Fix two nearly-self-cancelling errors:

1. The cross-sections in the magnetic reflecitivity kernel were misidentified as being in the order $(r^{--}, r^{-+}, r^{+-}, r^{++})$ (derivations and original docstrings indicate that they are in the opposite order)
2. The angle `Aguide` being sent to the calculation kernel was defined in the opposite sense to `EPS` in the derivations

Now, we interpret the outputs of the reflectivity kernel as being in the order $(r^{++}, r^{+-}, r^{-+}, r^{--})$, but then reverse them when returning from `refl1d.sample.reflectivity.magnetic_amplitude` in order to maintain backward compatibility with the order used in `PolarizedNeutronProbe` objects' cross-sections, and we explicitly pass `EPS = -Aguide` to the calculation kernel (again, for backward compatibility)

The end result is that 99% of users' models should continue to work without modification (as long as they do not have significant Zeeman-effect-induced splitting in the spin-flip scattering), and models that do show significant Zeeman effects should now work correctly.

The changes are documented in the `errata` section of the generated documentation, and examples of spin-flip scattering in high fields are provided in the `tutorials/validation` section of the docs (and validated against the `gepore` code developed by C. F. Majkrzak et al.)


pkienzle:
Having an inconsistent order of cross sections within our code base is likely to cause us confusion in the future. Could we instead treat this as a data reader problem, and reverse the order the cross sections are handled in the PNR probe?

Do we need the validation code in the released package? It seems to me that this is code we need to run once to generate the figures for the manual. Is there any reason for a user to run it? Do we need to run it during CI, or could we check against captured output from a single call to validation?

It we expect that a user will want to run gepore fortran code then we should document the process for installing gfortran on their machine.

bmaranville:
> Having an inconsistent order of cross sections within our code base is likely to cause us confusion in the future. Could we instead treat this as a data reader problem, and reverse the order the cross sections are handled in the PNR probe?
> 

I think a lot of user scripts interact with the xs by index - I am worried about breaking those, in a way where people wouldn't even know they were broken.  How do we mitigate that?

doc/errata/polarized_xs_order.rst
```diff
@@ -0,0 +1,60 @@
...
+book chapter by Majkrzak et al. [1]_, while simultaneously using
+the outputs of the magnetic calculation kernel in reversed order.
+
+Choosing an opposite convention for the `Aguide` angle causes the outputs of the calculation
+to be reversed when there is no splitting in the spin-flip cross sections, so these two
+choices effectively cancel each other out almost all the time.
+
+The exception is when there is significant splitting in the spin-flip cross sections,
+which occurs when there is a significant applied field **and** a significant projection of the
+magnetic moment perpendiculat to that applied field.  In this case, the energy corrections
```
acaruana2009, line 19: *perpendicular

acaruana2009 (approved):
I have gone through this briefly. Assuming that you have now checked this for consistency with the Gepore code (which from the comments you have), I think this is all good to go. 

We can use our test at the end of this cycle to confirm the spin states are doing the right thing, but I do not think the PR explicitly needs this confirmation.

bmaranville:
If run `python check_examples.py --chisq` on the current `master` branch (before the changes in this PR), I get this:
```
chisq 0.8396(50)
chisq 0.7872(100)
chisq 566.3524(87)
chisq 1.0121(25)
chisq 1.0121(25)
chisq 332.055(64)
chisq 0.996(30)
chisq 0.903(86)
chisq 0.0(50)e-3
chisq 0.857(24)
chisq 0.965(12)
chisq 2643.70(11)
chisq 6198.403(95)
chisq 0.975(24)
chisq 33.3532(98)
chisq 3.5731(24)
chisq 2.246(44)
chisq 0.7804(100)
chisq 0.846(88)
chisq 0.846(88)
chisq 0.00(83)e-3
chisq 9.496(29)
chisq 11.5714(21)
chisq 11.452(39)
chisq 11.5714(21)
```
If I run the same command after checking out the PR, I get:
```
chisq 0.8396(50)
chisq 0.7872(100)
chisq 566.3524(87)
chisq 1.0121(25)
chisq 1.0121(25)
chisq 332.055(64)
chisq 0.996(30)
chisq 0.903(86)
chisq 0.0(50)e-3
chisq 0.857(24)
chisq 0.965(12)
chisq 2643.70(11)
chisq 6198.403(95)
chisq 0.975(24)
chisq 33.3532(98)
chisq 3.5731(24)
chisq 2.246(44)
chisq 0.7804(100)
chisq 0.846(88)
chisq 0.846(88)
chisq 0.00(83)e-3
chisq 9.496(29)
chisq 11.5714(21)
chisq 11.452(39)
chisq 11.5714(21)
```
**NOTE**: they are identical, and this is what we expect for zero-field models (none of the checked examples is using a non-zero H field)

doc/conf.py
```diff
@@ -286,6 +287,34 @@ def download(path, filename):
...
+# -- header for Jupyter notebooks --------------------------------------------
+# This is processed by Jinja2 and inserted before each notebook
+nbsphinx_prolog = r"""
+{% set docname = 'doc/' + env.doc2path(env.docname, base=None)|string %}
+
+.. raw:: html
+
+    <div class="admonition note">
+      This page was generated from
+      <a class="reference external" href="https://github.com/reflectometry/refl1d/tree/main/{{ docname|e }}">{{ docname|e }}</a>.
```
pkienzle, line 299: `.../refl1d/tree/master/...` ?

Also, a couple of lines down uses `.../refl1d/blob/main/...` for the colab link.

pkienzle: Checking, the `tree` gets replaced by `blob` when I click on the link.

The link obviously doesn't work since the notebook is only in a branch. Maybe we should use the git hash instead of the branch name so that it looks at the notebook at the time the docs were built? Then it doesn't matter if it is master or main and it doesn't matter if the docs were built in a branch.

bmaranville: This was merged last week, so you should be able to try it out on the master branch.

I agree, if there's some way to use the commit or tag instead that would be useful.

PR #254, opened 2025-01-29
Fix webview reflectivity ylabels

bmaranville:


bmaranville:
The label for the $R \cdot Q^4$ plot was clearly wrong, and @hoogerheide verbally approved this change, so merging now...

PR #258, opened 2025-02-21
Use new Generic bumps FitProblem

bmaranville:
If `bumps.fitproblem.FitProblem[FitnessType]` is a Generic class with type parameter for the model type ( see https://github.com/bumps/bumps/pull/217 ) then we can use this when generating the JSON schema for refl1d models, as well as to assist type-checkers when building FitProblem objects with `refl1d.experiment.Experiment` model type.

This change allows us to completely remove the module `refl1d.bumps_interface.fitproblem`, which is then replaced with a FitProblem with the appropriate type parameter: `bumps.fitproblem.FitProblem[refl1d.experiment.Experiment]`


pkienzle (approved):
Removing code is good. I only inspected the changes, I didn't test them.

PR #260, opened 2025-02-26
Use git hash rather than branch name in the jupyter notebook header

pkienzle:


bmaranville:
If we only rebuild the docs on releases, and we also publish to pypi on every release, then the notebooks can generically use `%pip install refl1d` and it should "just work" because the most recent package corresponds to the current docs.

I think that also means we should also use the tag or hash in the link to pin the notebook, as a newer notebook being prepared for an upcoming release may not be compatible with the current release.

bmaranville (approved):

PR #265, opened 2025-03-19
Binary installers

bmaranville:
_Note that this is nearly the same PR as_ https://github.com/bumps/bumps/pull/221

Change infrastructure to build binary installers for MacOS and Windows rather than distributing compressed archives of the packed folders directly.

For linux, the compacted folder still seems to be a reasonable distribution method (tar is available, digital signatures not typically required)

This PR results in building

a MacOS dmg signed with our developer certificate, which contains:
a copy of the template in extra/platform_resources/macos/app_folder_template
bumps.app
bumps_shell.app
a shortcut to the /Applications folder, for ease of install
with the contents of the conda-packed bumps environment copied to bumps.app/Contents/Frameworks/python.framework/Resources/env
a Windows installer (made with NSIS)
also with the contents of the conda-packed environment, as the src of installed files
a shortcut which starts the webview server in a Powershell session
a shortcut which starts a Powsershell interactive session with the bumps environment activated (with micromamba), so that users can install more packages if they want.
Note that the Windows installer is not signed (we don't have a Windows certificate yet)

On all platforms, the "bumps-webview" command works as expected (in Windows, the conda-unpack command is run during installation, while on linux and macos the bumps-webview executable script already does the right thing)


extra/build_conda_packed.sh
```diff
@@ -1,78 +1,27 @@
...
 
-ENV_NAME="isolated-base"
-PYTHON_VERSION="3.11"
-PKGNAME="refl1d"
-SUBNAME="packed"
-OUTPUT="artifacts"
+# Definitions
+OUTPUT="conda_packed"
 SCRIPT_DIR=$(realpath $(dirname "${BASH_SOURCE[0]}"))
 SRC_DIR=$(dirname "$SCRIPT_DIR")
```
pkienzle, line 6: Should be quoted to protect against spaces in filenames.

extra/installer.nsi
```diff
@@ -0,0 +1,115 @@
...
+!include "LogicLib.nsh"
+!include "WinVer.nsh"
+!include "x64.nsh"
+
+;-------------------------------------------------------------------------------
+; Constants
+!ifndef PRODUCT_NAME
+    !define PRODUCT_NAME "Refl1D"
+!endif
+!define PRODUCT_DESCRIPTION "Bayesian Uncertainty Modeling for the Physical Sciences"
```
pkienzle, line 13: Refl1D reflectometry fitting package, not bumps.

extra/installer.nsi
```diff
@@ -0,0 +1,115 @@
...
+!define COPYRIGHT "Copyright © 2018 The Refl1D developers"
+!define PRODUCT_VERSION 1.0.0.1013
```
pkienzle, line 15: Refl1d release version and copyright should be supplied on the command line instead of hard coding them here.

.github/workflows/build-distributables.yml
```diff
@@ -29,26 +33,145 @@ jobs:
...
+        BUILD_DIR="${PACKAGE_NAME}_installer"
+        mkdir -p $BUILD_DIR
+        APP_FOLDER="$BUILD_DIR/$PRODUCT_NAME"
+        APP="$APP_FOLDER/${PACKAGE_NAME}.app"
+        PYTHON_FRAMEWORK="$APP/Contents/Frameworks/python.framework"
+        PYTHON_ENV="$PYTHON_FRAMEWORK/Resources/env"
+        ln -s /Applications "$BUILD_DIR/Applications"
+        cp -R -P "extra/platform_resources/macos/app_folder_template" "$APP_FOLDER"
+        cp -R -P "conda_packed" "$PYTHON_ENV"
+        plutil -replace CFBundleVersion -string "$VERSION" "$APP/Contents/Info.plist"
```
pkienzle, line 79: Do we also need to set CFBundleShortVersionString which is the user visible property?

https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleshortversionstring

Also, there appears to be an internal python framework. Should this be tagged with the python version number used?

What about the refl1d_shell app Info.plist?

.github/workflows/build-distributables.yml
```diff
@@ -29,26 +33,145 @@ jobs:
...
+        FULL_PACKAGE_NAME="$PACKAGE_NAME-$VERSION-Windows-$(uname -m)"
+
+        # clean up junk from conda-pack operation:
+        find conda_packed -name "*.conda_trash" -delete
+
+        conda install -y nsis
+
+        # create the installer:
+        PRODUCT_NAME="$PACKAGE_NAME-$VERSION"
+        $CONDA_PREFIX/NSIS/makensis.exe -DPRODUCT_NAME="$PRODUCT_NAME" extra/installer.nsi
```
pkienzle, line 138: Also add `$PRODUCT_VERSION` and `$COPYRIGHT`

.github/workflows/build-distributables.yml
```diff
@@ -29,26 +33,145 @@ jobs:
...
+        security unlock-keychain -p $KEYCHAIN_PASSWORD build.keychain
+        security import certificate.p12 -k build.keychain -P $MACOS_CERTIFICATE_PWD -T /usr/bin/codesign
+        security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k $KEYCHAIN_PASSWORD build.keychain
+        security find-identity -p codesigning
+
+        echo "Creating MacOS App"
+        # This assumes that conda-packed enviroment exists at "conda_packed"
+
+        VERSION=$(versioningit)
+        PRODUCT_NAME="$PACKAGE_NAME-$VERSION"
```
pkienzle, line 68: Can we put PRODUCT_NAME, PRODUCT_VERSION and COPYRIGHT in a common environment so they are shared across deployment platforms?

extra/build_conda_packed.sh
```diff
@@ -1,78 +1,27 @@
...
 
-cd $tmpdir
-# build the client
-$bindir/python -m refl1d.webview.build_client --cleanup
+cd $SCRIPT_DIR
+conda activate "$ISOLATED_ENV"
 
-version=$($bindir/python -c "import refl1d; print(refl1d.__version__)")
-mv "$tmpdir/$PKGNAME" "$tmpdir/$PKGNAME-$version"
+python -m pip install --no-input --no-compile "..[webview]"
```
pkienzle, line 20: Presumably this runs versioningit so that the version number in the bundle is correct?

extra/installer.nsi
bmaranville: True.  If we can just leave them out here that would be better.

.github/workflows/build-distributables.yml
bmaranville: The refl1d_shell.app is just a script that opens a shell with the environment from refl1d.app active - the shell app itself has no meaningful version.

It is only meaningful to put versions on the framework if we are planning to upgrade it or use it independently of the app.

extra/build_conda_packed.sh
bmaranville: Yes, I think versioningit is triggered during a pip install.  But I should check.

extra/installer.nsi
bmaranville: It looks like this info does show up if you right click on the installer and view Properties -> Details.  It won't be too much effort to add the correct information.

PR #276, opened 2025-05-02
Improve serialization of functional profiles

pkienzle:
Redo #219 using the generic function serialization in bumps.

Requires bumps serialize-functions branch for the tests to pass.

### DIFF
--- a/refl1d/sample/flayer.py
+++ b/refl1d/sample/flayer.py
@@ -1,14 +1,18 @@
+from dataclasses import dataclass
 import inspect
+from typing import Dict, Optional, Callable, Union
 
-from bumps.parameter import Calculation, Parameter, to_dict
+import numpy as np
 from numpy import asarray, broadcast_to, imag, real
 
+from bumps.parameter import Calculation, Constant, Parameter
+
 from refl1d import utils
-from .layers import Layer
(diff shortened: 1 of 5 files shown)


refl1d/sample/flayer.py
```diff
@@ -54,69 +59,111 @@ def linear(z, rhoL, rhoR):
...
+    def __setattr__(self, key, value):
+        # print(f"setting {key}")
+        if self.pars and key in self.pars and key not in self.__dict__:
+            if not isinstance(value, ValueProtocol):
+                raise TypeError("Can only assign parameter or expression to a parameter slot")
+            self.__dict__["pars"][key] = value
+        else:
+            super().__setattr__(key, value)
+
+    def __getattr__(self, key):
```
bmaranville, line 136: Wait... won't this block access to all the attributes that aren't in `pars`?

bmaranville (changes requested):
I am worried about the `__getattr__` override in FunctionalProfile... I don't see how you can access the regular attributes of the class with this in place.

bmaranville (approved):
I had forgotten the subtlety of when `__getattr__` is called... and the attribute-based access to pars is preserving the existing API - I have no further objections to this.

PR #278, opened 2025-05-05
Allow duplicate Q values in probes

bmaranville:
In stitched data (particularly in TOF data), it is not unusual to have multiple measurements defined for the same Q point, with different resolutions.

Currently, if a probe object is created that contains duplicate `Q` data points with different resolutions, refl1d raises an exception
(if it is a `refl1d.probe.QProbe` or a `refl1d.probe.PolarizedNeutronProbe`)

In part, I think this was in place to prevent having duplicate Q values in the calculated set of (calc_Q, calc_R) points that is used as the basis for doing the resolution correction on the modeled reflectivity (the `convolve_gaussian(...)` function in both the `refl1d.lib.c` and `refl1d.lib.python` backends has a divide-by-zero error if there are duplicate x (calc_Q) values in the inputs, **and** there are data points for which `sigma == 0` and `Q == calc_Q`).

On the other hand, the set of (calc_Q, calc_R) points is distinct from the set of data points, each of which has (Q, dQ, R, dR).  Notably, each data point has an associated dQ while the calc_Q points _do not_, so we can safely discard duplicate calc_Q points with no negative consequences.

In this PR:

- The guard is removed that explicitly raises an exception if duplicate Q are found in the set of (deduplicated) (Q, dQ) pairs
- np.unique is always applied to calc_Q before it is stored in the probe as `Probe.calc_Qo`

**NOTE** This guard has been raising an exception since 2011 (see e8730ff), so some testing to make sure that other things don't break for duplicate Q values in the data.

### DIFF
--- a/refl1d/probe/probe.py
+++ b/refl1d/probe/probe.py
@@ -856,7 +856,7 @@ class Probe(BaseProbe):
         idx = np.argsort(Q)
         self.calc_T = T[idx]
         self.calc_L = L[idx]
-        self.calc_Qo = Q[idx]
+        self.calc_Qo = np.unique(Q[idx])
 
         # Only keep the scattering factors that you need
         self.unique_L = np.unique(self.calc_L)
@@ -1398,7 +1398,7 @@ class QProbe(BaseProbe):
         self.R = R
         self.dR = dR
         self.unique_L = None
-        self.calc_Qo = self.Q
+        self.calc_Qo = np.unique(self.Q)
         self.name = name
         self.filename = filename
         self.resolution = resolution
@@ -1425,7 +1425,7 @@ class QProbe(BaseProbe):
         rng = numpy.random.RandomState(seed=seed)
         extra = rng.normal(self.Q, self.dQ, size=(n - 1, len(self.Q)))
         calc_Q = np.hstack((self.Q, extra.flatten()))
-        self.calc_Qo = np.sort(calc_Q)
+        self.calc_Qo = np.unique(calc_Q)
 
     oversample.__doc__ = Probe.oversample.__doc__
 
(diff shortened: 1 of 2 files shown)


glass-ships (approved):

acaruana2009 (approved):

refl1d/probe/probe.py
```diff
@@ -856,7 +856,7 @@ def _set_calc(self, T, L):
         idx = np.argsort(Q)
         self.calc_T = T[idx]
         self.calc_L = L[idx]
-        self.calc_Qo = Q[idx]
+        self.calc_Qo = np.unique(Q[idx])
```
pkienzle, line 859: This breaks the link between the length of `calc_Qo` and the lengths of `calc_T` and `calc_L`. That it is working at all means that `calc_T` and `calc_L` are not used anywhere and can be removed.

pkienzle: Okay, not strictly true. It is used for theta offset when generating a new calc_Q.

Perhaps we should drop the unique, and simply allow the occasional duplicate.

Theta offset does not protect against duplicate Q values appearing, so the resolution code has to be robust against duplicates.

Separate issue: I believe resolution expects non-decreasing Q. We should therefore be sorting the Q values returned from the Q property. Perhaps we could cache the unique Q for the current theta_offset setting.

refl1d/probe/probe.py
```diff
@@ -1395,14 +1396,14 @@ def __init__(
...
-        self.calc_Qo = self.Q
+        self._calc_Q = np.unique(self.Q)
         self.name = name
         self.filename = filename
         self.resolution = resolution
 
     @property
     def calc_Q(self):
-        return self.calc_Qo if not self.back_reflectivity else -self.calc_Qo
+        return self._calc_Q if not self.back_reflectivity else -self._calc_Q
```
pkienzle, line 1406: Did you check whether the resolution calculation can handle decreasing instead of increasing Q values? Or is resolution doing its own sort just to be safe.

pkienzle (approved):

refl1d/probe/probe.py
bmaranville: I'm not sure.  That particular logic has been in place for a long time (returning -calc_Q for back_reflectivity=True) and hasn't changed with this PR - but we should definitely check it!

bmaranville: Yes, in `BaseProbe.apply_beam` the calc_Q is negated after the reflectivity is calculated if back_reflectivity is True before running the resolution correction: https://github.com/reflectometry/refl1d/blob/master/refl1d/probe/probe.py#L256

PR #284, opened 2025-05-19
pass socket object as prop to main app

bmaranville:
See https://github.com/bumps/bumps/pull/268

(do not merge until that is published)

### DIFF
--- a/refl1d/webview/client/src/main.js
+++ b/refl1d/webview/client/src/main.js
@@ -1,12 +1,21 @@
 import "bootstrap/dist/css/bootstrap.min.css";
 import { computed, createApp } from "vue";
-import { file_menu_items, fileBrowser, shared_state, socket } from "bumps-webview-client/src/app_state";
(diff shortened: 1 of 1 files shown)


refl1d/webview/client/src/main.js
```diff
@@ -1,12 +1,21 @@
...
 
 const name = "Refl1D";
+const urlParams = new URLSearchParams(window.location.search);
+const singlePanel = urlParams.get("single_panel");
+const sio_base_path = urlParams.get("base_path") ?? window.location.pathname;
+const sio_server = urlParams.get("server") ?? "";
+
+const socket = io(sio_server, {
+  path: `${sio_base_path}socket.io`,
+});
```
pkienzle, line 18: This duplicates code in bumps client main.js. Can it be moved into a bumps library function callable from refl1d?

bmaranville: not easily...

pkienzle (approved):

PR #290, opened 2025-05-23
install the fitplugin to bumps before starting cli

bmaranville:
This is needed for registering the serialized model migrations, as well as the specialized refl1d loader.

Currently the cli and webview interfaces start just fine without this PR, but older serialized models are not being migrated and therefore can't be opened.


bmaranville:
The work on refactoring the plugin infrastructure to use entry points is largely complete, but testing etc. has been put off until after bumps 1.0, so I think we should do this (extremely minimal) PR in order to get refl1d working in the meantime, as this is a bugfix.

pkienzle (approved):

PR #293, opened 2025-05-30
Fix plot option for oversampling helper tool

bmaranville:
pass plot option to the oversampling checker (currently broken)


bmaranville:
lol yes, they are, but I was calling them positionally, and added a new one `seed` right in the middle, so the call from `main()` was not passing `plot` in the right place.

bmaranville:
It's been broken (no plot shown) for a long time, but no one noticed until today

PR #300, opened 2025-06-17
bugfix: cli flags converted to dashes but not updated in distributables

bmaranville:


pkienzle (approved):
The code looks fine but I didn't test it.

PR #301, opened 2025-06-18
Orso sample loader

bmaranville:
Adds a translator for creating a refl1d sample from a sample model in the orso specification.

- [ ] Add example ORSO sample definition


bmaranville:
It seems likely that we'll have to clean up the `refl1d.sample.material.Mixture` class before we use this, because some of the most prominent example ORSO sample definitions use the `Composit` class which maps most directly onto `Mixture`.

PR #302, opened 2025-06-23
disable prettier-plugin-jsdoc until it is fixed

bmaranville:


bmaranville:
https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/245

bmaranville:
We also have very little (or no?) jsdoc content in this project, so we may be able to drop it permanently.

PR #309, opened 2025-07-15
toggle background in data vs theory curve

hoogerheide:
Pathological plotting issues are common when plotting reflectivity data, particularly when negative values of `probe.background` are present. For example, in a simple logarithmic plot, the theory curve is unreadable:
image

The problem is especially apparent when normalizing by the Fresnel reflectivity:
image

Even linear Fresnel plots have this problem:
image

Both problems arise because the data are fixed and the theory curve moves around the data as the background changes.

This PR creates a toggle ($R_{bkg}$) that allows the theory curve to be fixed to zero reflectivity, while the background is applied to the data.
image

This solves both problems:
image
(Note that additional data are now missing from this plot, but there were already negative data missing in the original so this is not a new problem.)
image

For Fresnel plots, the missing data issue is easily solved by switching to a linear scale, which now looks quite reasonable:
image

### DIFF
--- a/refl1d/webview/client/src/components/DataView.vue
+++ b/refl1d/webview/client/src/components/DataView.vue
@@ -16,7 +16,7 @@ const chisq_str = ref("");
 const log_y = ref(true);
 const log_x = ref(false);
 const show_resolution = ref(true);
-const subtract_background = ref(true);
+const apply_corrections = ref(true);
 const show_residuals = ref(false);
 
 const props = defineProps<{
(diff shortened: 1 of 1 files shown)


hoogerheide:
At the moment the toggle is set to `true`, meaning it will automatically shift the data (rather than the theory curve) by the background. No opinion on which way should be default; to maintain continuity of behavior, this should probably default to `false`

pkienzle:
Need to make sure that the plots in probe.py are doing the right thing (they are still used for --export, though probably not with the fresnel view unless you go through some effort).

hoogerheide:
These are good changes. I like applying the scale correction to the data in addition to the background subtraction.

PR #310, opened 2025-07-16
Add oversampled_regions to PolarizedNeutronProbe and PolarizedQProbe

bmaranville:
In the earlier PR adding `Probe.oversampled_regions` and `QProbe.oversampled_regions` no similar capability was added to `PolarizedNeutronProbe` or `PolarizedQProbe`

- `oversampled_regions: List[OversampledRegion]` attribute is added to `PolarizedNeutronProbe`
- `oversampled_regions` attribute is added to `PolarizedQProbe`
- these are properly acted on during `PolarizedNeutronProbe._calculate_union()` (which is inherited by `PolarizedQProbe` as a method)

In addition, an implementation of extending
Now, all the probe classes should have `oversampled_regions` implemented.

### DIFF
--- a/refl1d/probe/probe.py
+++ b/refl1d/probe/probe.py
@@ -1871,12 +1871,16 @@ def _get_oversampled_values(V_in, dV_in, oversampling, oversampling_seed, oversa
 
 def _get_normal_oversampling_points(V_in, dV_in, n, seed):
     rng = numpy.random.RandomState(seed=seed)
+
+    # TODO: allow extending the range of V_in to support resolution
     # prepend values out to -3*dV_in and +3*dV_in
-    append_points = np.arange(1.0, 4.0)
-    prepend_points = -append_points[::-1]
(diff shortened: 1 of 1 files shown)


bmaranville:
When you calculate R, you get all 4 cross sections whether you want to or not, so there's no way to improve efficiency by trying to reduce the calculations for the spin-flip data.

pkienzle (approved):
We don't have a good test suite for these operations. I verified that PolarizedNeutronProbe.oversampled_regions does something with the following code in `doc/examples/spinvalve/n101G.py`:
```python
from refl1d.probe.probe import OversampledRegion
probe.oversampled_regions = [OversampledRegion(Q_start=0.009, Q_end=0.011, n=50)]
```
PolarizedQProbe inherits from PolarizedNeutronProbe, so I didn't test it directly.

If this is the expected way to add oversampling regions to a loaded probe then we need to add OversampledRegion to `probe/__init__.py` and `names.py`.

hoogerheide (approved):
I also don't have a good way to test this since I don't have any polarized beam models. But it looks reasonable to me.

pkienzle (approved):

PR #331, opened 2025-09-25
apply data_range to all columns, including T, dT, L, dL

bmaranville:
Fixes an issue where if a dataset with more than 4 columns is loaded (including T, dT, L or dL) and the `data_range` flag is used, then the Q, R, dR, dQ columns are trimmed to the data range, but the additional columns are not.

### DIFF
--- a/refl1d/probe/data_loaders/load4.py
+++ b/refl1d/probe/data_loaders/load4.py
@@ -353,6 +353,12 @@ def _data_as_probe(
     data_T = fetch_key("angle", T)
     data_L = fetch_key("wavelength", L)
 
+    # Apply index to T and L if they are arrays
+    if data_T is not None and hasattr(data_T, "__getitem__"):
+        data_T = data_T[index]
+    if data_L is not None and hasattr(data_L, "__getitem__"):
+        data_L = data_L[index]
+
     # If one of T and L is missing, reconstruct it from Q
     if data_T is None and data_L is not None:
         data_T = QL2T(data_Q, data_L)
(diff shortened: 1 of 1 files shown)


pkienzle (approved):
Ick. Can we instead force the arity of T,L,dT,dL to match R?

bmaranville:
You mean before we apply the slice?  Is there a compact way to do that?  It's tricky because they can be None, too.  I didn't check what would happen if they were a scalar - if the  load would still work.

acaruana2009 (approved):

glass-ships (approved):

pkienzle:
An alternative fix is to use the following in `fetch_key`:
```python
np.asarray(v)[index] if hasattr(v, '__getitem__') else v
```

And maybe rename the function to `fetch_and_index`

refl1d/probe/data_loaders/load4.py
```diff
@@ -87,26 +87,46 @@ def get_key(orso_name, refl1d_name, refl1d_resolution_name):
                 (i for i, c in enumerate(columns) if getattr(c, "physical_quantity", None) == orso_name),
                 None,
             )
+            resolution_index = None
```
pkienzle, line 90: get_key() does not need to be defined in the for loop. It picks up the local scope regardless of where things are defined, so long as it is defined before it is used.

refl1d/probe/data_loaders/load4.py
```diff
@@ -87,26 +87,46 @@ def get_key(orso_name, refl1d_name, refl1d_resolution_name):
...
         get_key("incident_angle", "angle", "angular_resolution")
         get_key("wavelength", "wavelength", "wavelength_resolution")
 
+        # convert error columns from FWHM to sigma if they are present
+        # and need conversion
+        if len(columns) >= 3:
+            # third column is always uncertainty on reflectivity
+            dR_col = columns[2]
+            if dR_col.value_is == "FWHM":
+                data[:, 2] = FWHM2sigma(data[:, 2])
```
pkienzle, line 123: This modifies the entry in place as returned by the orso/nexus reader. That's probably okay, but since you are copying the data anyway on line 130, you may as well move the copy to line 62 so that you are operating on your own copy of the data. You might also force the data type to be float64 when you create the copy.

I'm a little concerned that ΔR is potentially being scaled by FWHM2sigma. Is there any risk that the ΔR column is mislabeled as FWHM, e.g., because it is defaulting to FWHM when the value isn't specified? This would be happening in the orso/nexus readers.

PR #332, opened 2025-09-26
Print version number when running webview

glass-ships:
CIS requested feature to print refl1d and bumps version numbers when running the webview client,
useful for easily identifying which version is being run in the event some issues are noticed in deployment.

### DIFF
--- a/refl1d/webview/server/cli.py
+++ b/refl1d/webview/server/cli.py
@@ -6,6 +6,7 @@ import sys
 import asyncio
 from pathlib import Path
 
+from bumps import __version__ as bumps_version
 from bumps.webview.server import cli
 
 from . import api  # uses side-effects to register refl1d functions
@@ -32,6 +33,12 @@ def main():
         del sys.argv[1]
         run_errors()
     else:
+        print(f"""
+{"=" * 50}
+Refl1d version:\t{__version__}
+with Bumps:\t{bumps_version}
+""")
+
         cli.plugin_main(name="refl1d", client=CLIENT_PATH, version=__version__)
 
 


pkienzle:
bumps.webview.server.cli seems like a better place for this. Something like:
```diff
diff --git a/bumps/webview/server/cli.py b/bumps/webview/server/cli.py
index 6c25456f..f80a3fa9 100644
--- a/bumps/webview/server/cli.py
+++ b/bumps/webview/server/cli.py
@@ -409,12 +409,14 @@ def get_commandline_options(arg_defaults: Optional[Dict] = None):
         help="display logging to console",
         default="warn",
     )
-    # TODO: show version numbers for both refl1d and bumps?
+    # Show version numbers for both refl1d and bumps
+    from bumps import __version__
+    bumps_tag = f" bumps {__version__}" if api.state.app_name != "bumps" else ""
     misc.add_argument(
         "-V",
         "--version",
         action="version",
-        version=f"%(prog)s {api.state.app_version}",
+        version=f"%(prog)s {api.state.app_version}{bumps_tag}",
     )
 
     # TODO: restructure so that -b -s -r --webview override each other
```

glass-ships:
There's also precedent in ORNL apps to print out the version when starting the GUI (separate from having a `--version` CLI flag, this PR looks like:
```
> refl1d -x

==================================================
Refl1d version: 1.0.0rc0.post3+geae6af51
with Bumps:     1.0.3rc2

======== Running on http://127.0.0.1:5148 ========
(Press CTRL+C to quit)
```
for example. 

switching to a logger rather than print seems reasonable, but because of how bumps handles logging, the output can get quite strange. 

if for example, I import `bumps.webview.client.logger` and use that, it only shows up if i log at the `warning` level, but if i use the standard python logger, that configuration seems to interfere with bumps's logging so it ends up looking like: 
```
> refl1d -x
WARNING:root:
==================================================
Refl1d version: 1.0.0rc0.post3+geae6af51
with Bumps:     1.0.3rc2

INFO:webview:BumpsOptions(filename=None, args=None, fit_options={}, resume=False, session=None, read_session=None, write_session=None, serializer='dataclass', no_auto_history=False, path=None, use_persistent_path=False, reload_export=None, pars=None, simulate=False, simrandom=False, shake=False, noise=5, seed=0, chisq=False, export=None, trace=False, parallel=0, mpi=False, mode='edit', headless=True, external=False, port=0, hub=None, convergence_heartbeat=False)
======== Running on http://127.0.0.1:5148 ========
(Press CTRL+C to quit)
^CINFO:webview:shutdown task
INFO:webview:killing...
INFO:webview:shutdown complete
INFO:webview:cleanup task
INFO:webview:webserver shutdown tasks complete
```

bmaranville:
This seems to work for both bumps and refl1d, extending Paul's suggestion a bit:
```diff
diff --git i/bumps/webview/server/cli.py w/bumps/webview/server/cli.py
index 5c71d756..7068373e 100644
--- i/bumps/webview/server/cli.py
+++ w/bumps/webview/server/cli.py
@@ -395,7 +395,7 @@ def get_commandline_options(arg_defaults: Optional[Dict] = None):
         "-V",
         "--version",
         action="version",
-        version=f"%(prog)s {api.state.app_version}",
+        version = branding()
     )
 
     # TODO: restructure so that -b -s -r --webview override each other
@@ -865,6 +865,16 @@ def plugin_main(name: str, client: Path, version: str = ""):
     api.state.client_path = client
     main()
 
+def branding():
+    from bumps import __version__ as bumps_version
+
+    output = f"Bumps version: {bumps_version}\n"
+    if api.state.app_name != "bumps":
+        output += f"{api.state.app_name} version: {api.state.app_version}\n"
+    output += f"Python version: {sys.version.splitlines()[0]}\n"
+    output += f"Platform: {sys.platform}"
+    return output
+
 
 def main(options: Optional[BumpsOptions] = None):
     # TODO: where do we configure matplotlib?
@@ -909,6 +919,7 @@ def main(options: Optional[BumpsOptions] = None):
         # api.state.rank = ""
 
     if webview and is_controller:  # gui mode
+        print(branding())
         start_from_cli(options)
     else:  # console mode
         run_batch_fit(options)
```

pkienzle:
start_bumps/display_bumps are in one of the as yet unapproved PRs, in webserver.py:
```python
# CRUFT: aliases for deprecated names
display_inline_jupyter = display_bumps
start_bumps_server = start_bumps
```

PR #333, opened 2025-09-30
adding release notes for v1.0.0

bmaranville:


pkienzle (approved):

bmaranville:
These notes were used in the github release - I don't know if we want to keep them in the git tree as well.  As-is, this is just another markdown file in the repo, but if we keep it then we might want to include it in the sphinx docs.

bmaranville:
@pkienzle that seems reasonable - but only if we can use markdown for the changelog, so we can cut and paste back and forth between release notes on github and the changelog.  Does that sound ok?

PR #345, opened 2025-11-13
compensate for missing background and intensity values from API

bmaranville:
Currently the display of reflectivity for new models from SimpleModelBuilder is broken, because of a missing intensity and background parameter (since some recent changes enhancing the display when those parameters are included in the model)

This PR changes the logic in the reflectivity panel, adding a default value for intensity (1.0) and background (0.0) if none is provided by the model.


bmaranville:
Since this is broken in the current release of refl1d, I think it would be good to fix and then make a new release soon!

bmaranville:
...marginally better to add it to builder logic, but definitely much more complex of a fix.

EDIT: ok, now I'm actually stumped, because looking at the SImpleBuilder code it _does_ attach an intensity and background, or at least it tries to.  Some more investigation is probably warranted.

Issue #341, opened 2025-10-30
Ability to save out the model builder state in webview to a .py script - for teaching purposes


acaruana2009:
It would be useful if there was an option to save out the state of the model builder to a .py script (similar to how this can be done in the NIST reflectivity calculators). 

This could be kept to quite a limited scope, and we could decide some specific ways we would convert from for example a defined slab in the model builder to what it would look like in code - possibly could add some generic annotations as well if it makes sense.

The main aim of this would purely be for teaching purposes and lower the barrier to entry for scripting.

PR #346, opened 2025-11-13
Add utility for converting simple models to python scripts

bmaranville:
This PR adds a capability of converting simple models to annotated python scripts.  

Supports all models that can be built with the current SimpleModelBuilder.

Attempts to use load4 if a Probe was loaded from a data file.

The GUI interface is not tested.

Addresses #341

### DIFF
--- a/refl1d/webview/client/src/components/SimpleBuilder.vue
+++ b/refl1d/webview/client/src/components/SimpleBuilder.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { computed, onMounted, ref, shallowRef } from "vue";
 import type { ComputedRef } from "vue";
+import { fileBrowser } from "bumps-webview-client/src/app_state";
 import type { AsyncSocket } from "bumps-webview-client/src/asyncSocket";
 import { v4 as uuidv4 } from "uuid";
 import { dqIsFWHM } from "../app_state";
@@ -262,6 +263,24 @@ function setQProbe() {
   sendModel();
 }
 
+async function exportModelScript() {
+  if (fileBrowser.value) {
+    const settings = {
+      title: "Export Model Script to File",
+      callback: (pathlist: string[], filename: string) => {
+        props.socket.asyncEmit("export_model_script", pathlist, filename);
+      },
+      show_name_input: true,
+      name_input_label: "Filename",
+      require_name: true,
+      show_files: false,
+      chosenfile_in: "model.py",
+      search_patterns: [".py"],
+    };
+    fileBrowser.value.open(settings);
+  }
+}
+
 onMounted(() => {
   fetchModel();
 });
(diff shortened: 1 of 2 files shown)


pkienzle:
Can we delegate scriptify to the underlying data structures rather than maintaining a parallel code base that must be kept in sync with any changes to experiment, probe and sample?

For now they only need to support the features required for the GUI builder, not the full set of controls available to the scripted model interface.

bmaranville:
No, it's not really possible to make a meaningfully commented and structured script "inside-out", by asking the model to render itself as a script: that would end up looking like a nested class instantiation (it would look a lot like the JSON serialization), which would be much less useful as a starting point for a script.

bmaranville:
I have tested that this works, and minimally meets the requirement of being able to export a problem script for users to work with.

Currently if you load data into a model from the GUI, it will add `probe = load4(<full absolute path to file>)` for the file that was loaded, which is probably not what you want for portable models.  Not sure what the best solution is.

It references parameters by where they are in the model, which is not ideal, but it _does_ also include an inline comment with the parameter's full name, which can be helpful I hope!  Here is an example output:

```python
from refl1d.names import *

################
# Experiment_1 #
################


probe_1 = load4("/home/bbm/dev/refl1d-modelbuilder/NbAl/BK_NbAl_150mT-dream/BK_NbAl_150mT-refl.datA")

slabs_1 = [
    Slab(name="Si", thickness=0.0, interface=1.0, material=SLD(name="Si", rho=2.07, irho=0.0), magnetism=None),
    Slab(name="sld", thickness=250.0, interface=1.0, material=SLD(name="sld", rho=2.5, irho=0.0), magnetism=None),
    Slab(name="Vacuum", thickness=0.0, interface=0.0, material=SLD(name="Vacuum", rho=0.0, irho=0.0), magnetism=None),
]

sample_1 = Stack(slabs_1)

#####################
# Sample Parameters #
#####################

# slabs_1[0].thickness.range(-0.1, 0.1) # Si thickness
# slabs_1[1].thickness.range(125.0, 375.0) # sld thickness
# slabs_1[2].thickness.range(-0.1, 0.1) # Vacuum thickness

# slabs_1[0].interface.range(0.5, 1.5) # Si interface
# slabs_1[1].interface.range(0.5, 1.5) # sld interface
# slabs_1[2].interface.range(-0.1, 0.1) # Vacuum interface

# slabs_1[0].material.rho.range(1.035, 3.1049999999999995) # Si rho
# slabs_1[1].material.rho.range(1.25, 3.75) # sld rho
# slabs_1[2].material.rho.range(-0.1, 0.1) # Vacuum rho

# slabs_1[0].material.irho.range(-0.1, 0.1) # Si irho
# slabs_1[1].material.irho.range(-0.1, 0.1) # sld irho
# slabs_1[2].material.irho.range(-0.1, 0.1) # Vacuum irho



# === Experiment Arguments ===
# a model (Experiment) object is initialized with these arguments:
# with the following arguments:
#  - sample (required) is a Stack object that contains the layers of the sample
#  - probe  (required) is a Probe object that contains the probe information
#  - dz is the step size in Angstroms to be used for rendering the profile
#  - step_interfaces is a boolean that determines whether to use the
#    Nevot-Croce approximation for roughness (True) or to use the
#    microslabbed interfaces (False)
#  - dA is an aggregation constant for combining smooth regions of the profile
#    prior to calculating R (greatly speeds up fits)

experiment_1 = Experiment(sample=sample_1, probe=probe_1, dz=1.0, step_interfaces=False, dA=1.0)

##############
# FitProblem #
##############

problem = FitProblem([experiment_1])
```

bmaranville:
We should probably handle the case where a parameter doesn't have user-defined bounds, and use the limits instead or something like that.

PR #347, opened 2025-11-13
always send the background and intensity values to plotter

bmaranville:
When no data was attached to problem, API was not sending background and intensity values with `get_plot_data`.

This PR fixes that.  Replaces now-unneeded #345


bmaranville:
I tested... now models built with Builder tab properly show in Reflectivity tab, even when no data is loaded.

PR #353, opened 2025-12-04
Switch to Oxlint/Oxfmt

glass-ships:
- Switches to using [oxlint](https://oxc.rs/) for frontend linting
- Addresses a couple minor linting and type-checking issues
- Modify the frontend lint pre-commit hook to check for bun or npm rather than only using npm

Same motivation as #342, but oxlint is by the Vue team and actually supports Vue style files, which biome currently does not support. 

Full compatibility with eslint and prettier rules (no new or missing linting errors / formatting changes were made in running the new tool(s).

Will follow up with an analogous bumps PR if this is received well


bmaranville (approved):

glass-ships:
nice, you got it haha. 
before we merge this, actually, I did just find out that biome 2.3 released which does also now include full support for Vue. I'm interested in updating that branch and seeing how we feel about one tool vs. the other

bmaranville:
It seems like both `oxc` and `biomejs` are missing the ability to deeply analyze the template part of Vuejs components like the eslint plugin does - is that something we want to give up? (they analyze the JS/TS but not the template?)

glass-ships:
oh, one thing we may want to bring over from the biome PR is that `DataView` global property being shadowed

bmaranville:
I feel like there was never an overwhelming determination that this was better or worse than biomejs.  To be honest I forgot about this PR when I was working over in the bumps repo, and I needed a formatter that worked when prettier was breaking again, and there was already a `biomejs` PR that I could start from in bumps.

glass-ships:
Yeah, this seemed promising since it comes from the Vue team themselves, but like biome, it still doesn't support linting the <template> section, which doesn't really make it any better than biome, and we've put the work into biome already. 

I'd say we can just close this for now, and whenever one or the other wins the race to template linting, we can revisit the comparison

PR #354, opened 2025-12-05
Biomejs again, with setting more similar to before

bmaranville:
Just like #342 but using settings that make less aggressive changes from before

- allows multiple attributes on the same line : `"vue/max-attributes-per-line": ["error", { singleline: { max: 5, } }]`
- doesn't try to adjust self-closing (or void) elements: `"vue/html-self-closing": [ "error", { html: { void: "any", normal: "any", }, }, ],`


bmaranville:
closing in favor of #372 ... will revisit settings here when we can use biomejs for lint as well.

PR #366, opened 2026-03-24
update to vite 8 (fast client builds)

bmaranville:
Vite 8 was recently released, and it uses `rolldown` which is a much faster bundling engine.  As a result build times for the client are about 1s, compared to 8-9 seconds before.  Makes client development more pleasant.


pkienzle (approved):
Simple change. Seems to work.

PR #373, opened 2026-05-06
Shift all profiles so aligned interface is at relative z = 0 in profile uncertainty

bmaranville:
This PR changes the way profile uncertainty is displayed.

- In the plot, the location of the "alignment interface" is set to be z=0 for all profiles
- Contours are plotted relative to this zero
- X-axis shows "relative z (A)" instead of "z"

This has the (desired) effect of co-aligning all models to the same relative z, even if the number of layers is different, e.g. for the models shown in #355 

Fixes #355

**NOTE**
I don't have a good test case for this handy.  If anyone can share a session file with a dream fit of a multi-model problem with a different number of layers in the models, I would appreciate it!


mdoucet (approved):

pkienzle (approved):
Tested on a single profile. It indeed puts z=0 at the alignment index.

PR #376, opened 2026-06-06
fix(webview): prevent network-drive stalls in file browser dirlisting

pquarterman17:
## Problem

The webview file browser hangs for 30+ seconds when navigating network drives.

The root cause is in `bumps.webview.server.api.get_dirlisting` (which refl1d uses). For every directory entry it:

- counts each subdirectory's contents with `len(list(p.glob("*")))` — a full glob *per subfolder*, and
- does a redundant `p.exists()` check on entries just returned by `iterdir()`,

all **synchronously on the asyncio event loop**. On a high-latency network filesystem each glob is a round trip, so a folder with many subfolders blocks the server for tens of seconds and the whole UI freezes.

## Fix

Override `get_dirlisting` in `refl1d/webview/server/api.py` with a version that:

- **skips per-subfolder content counting** (`size = 0` for directories — the count isn't shown in the UI anyway),
- **drops the redundant `exists()` check**,
- runs the directory walk in a **worker thread** (`asyncio.to_thread`) so it never blocks the event loop, and
- adds a **10 s timeout** that returns a friendly error instead of hanging when a path is unreachable.

The return value is **identical in shape** to bumps' `get_dirlisting` (`drives`, `pathlist`, `subfolders`, `files`), so it is a drop-in replacement for the existing frontend.

## Notes / discussion

- The slow code lives in **bumps**, not refl1d — this PR overrides it on the refl1d side. If you'd rather fix it upstream in `bumps.webview.server.api.get_dirlisting` directly (benefiting every bumps-webview consumer, and avoiding a duplicated function here), I'm glad to redirect this there instead.
- Minor behavioral change: when a path does not exist, this returns `{"error": ...}` rather than bumps' "notify + fall back to cwd". Trivial to match bumps' behavior if preferred.
- `os.listdrives()` is Windows / Python 3.12+ and is guarded by `hasattr` exactly as in bumps, so non-Windows platforms get `drives = []`.

## Testing

- `py_compile` clean; confirmed the return shape matches bumps' `get_dirlisting`.
- Developed against a real network-drive stall on Windows: navigation that previously hung ~30 s now returns promptly, and unreachable paths time out at 10 s with an error instead of freezing the UI.

### DIFF
--- a/refl1d/webview/server/api.py
+++ b/refl1d/webview/server/api.py
@@ -1,4 +1,5 @@
 import asyncio
+import os
 from copy import deepcopy
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


bmaranville:
This seems like a good change.  You are correct that it should be in bumps instead - I'll have a close look at it next week

pquarterman17:
I'll see if I can make a bumps version of this PR to more directly fix; I have a couple other PRs for bumps itself anyway, this one was just a particular stickler for me til implementing this on the refl1d side

pquarterman17:
As offered in the PR description, the underlying fix is now submitted upstream as bumps/bumps#442, which repairs `get_dirlisting` in place (off-loop directory walk, no per-subfolder glob, 10 s timeout) while preserving bumps' existing notify-and-fall-back-to-cwd behavior for missing paths. If that lands, this override becomes unnecessary — happy to either close this PR in favor of the bumps fix, or keep it as a stopgap until refl1d picks up a bumps release containing it.

pkienzle:
Fix it in bumps.

pquarterman17:
closing as fix in bumps is in progress

Issue #339, opened 2025-10-30
Repeated magnetic layers not implemented


acaruana2009:
Currently it is not possible to define a magnetic multilayer in the same way you can specify a non-magnetic multilayer. E.g.:

```python
Si = Material(formula="Si")
Ni = Material(formula="Ni[58]")
Si_ml = Material(formula="Si")


Si_sub = Slab(material=Si, thickness=0, interface=5)
Ni_layer = Slab(material=Ni, thickness=100, interface=5)
Si_layer = Slab(material=Si_ml, thickness=100, interface=5)

# Sample construction/Stack

ML = (Ni_layer(magnetism=Magnetism(rhoM=2.0, interface_above=5, interface_below=5, name="Ni Layer Sample 1"))
      | Si_layer)

sample = (Si_sub
          | ML*2
          | air
          )
```
Raises: `NotImplementedError("Repeated magnetic layers not implemented")` from line 162 in refl1d/profile.py

The work around is to define the magnetism in the Slab instantiation, and define the multilayer as a list/tuple: `ML = (Ni_layer, Si_layer)`
This has the disadvantage that if you are doing multi-fits (say multiple fields), unlinking the magnetism between fits becomes tedious.

PR #377, opened 2026-06-10
Enable repeated magnetic layers

bmaranville:
Addresses #339 

In this PR, special handling is added in the `Repeat.render()` method, so that if the stack is magnetic, it renders it `repeat` times sequentially, while still using the more optimized `slabs.repeat()` method for non-magnetic stacks.

It also fixes a small bug in identifying when a Stack is near the substrate, during `Stack._render_magnetic`, since `i==0` will be true at the start of any `Stack` enumeration, and not just the first one.  It is safer to check if any slabs have been created yet (`len(slabs) == 0`)

Adds a test which generates a multilayer Sample in two ways, e.g. 

- `Si | (Ni(magnetism=...), Si) * N | air`, which _doesn't_ create a Repeat class, vs. 
- `Si | (Ni(magnetism=...) | Si) * N | air` which _does_ create a `Repeat`, and which currently breaks because the `Repeat.render` method raises an exception if the stack is magnetic.

Then compares the profiles and reflectivities of the samples for the same `Probe`.

Model evaluation time is identical between the two methods.

There is a subtle difference between `(Ni|Si)*N -> Repeat(Ni|Si, N=N)` and `(Ni, Si)*N -> (Ni, Si, ..., Ni, Si)` in that the interface between the multilayer and the surround uses `Repeat.interface` in the first case and `Si.interface` in the second case.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/refl1d/sample/layers.py
+++ b/refl1d/sample/layers.py
@@ -370,7 +370,7 @@ class Stack(Layer):
                 anchor = slabs.thickness() + magnetism.dead_below.value
                 s_below = (
                     np.nan
-                    if i == 0
+                    if len(slabs) == 0
                     else magnetism.interface_below.value
                     if magnetism.interface_below is not None
                     else slabs.surface_sigma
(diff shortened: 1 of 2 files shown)


bmaranville:
Note that this leaves the door open to optimizing the rendering of magnetic stacks in the future, with zero change to the API.

pkienzle (approved):
Code looks good and tests pass. I updated the tests so I could play with the models with sliders in webview.

acaruana2009 (approved):

PR #387, opened 2026-08-05
turn on auto-tagging in Experiment

bmaranville:
This PR 

- sets the default value of `auto_tag` in the Experiment initializer to `True`
- changes the first tag applied to `Probe` parameters from "probe" to "instrument"
- adds "nuisance" tag to `Probe` parameters

### DIFF
--- a/refl1d/deprecated/freeform.py
+++ b/refl1d/deprecated/freeform.py
@@ -179,7 +179,7 @@ class FreeInterface(Layer):
 
     def parameters(self):
         return {
-            "interface": self.interface.parameters(),
+            "interface": self.interface,
             "below": self.below.parameters(),
             "above": self.above.parameters(),
             "dz": self.dz,
--- a/refl1d/deprecated/freeform.py
+++ b/refl1d/deprecated/freeform.py
@@ -179,7 +179,7 @@ class FreeInterface(Layer):
 
     def parameters(self):
         return {
-            "interface": self.interface,
+            "interface": self.interface.parameters(),
             "below": self.below.parameters(),
             "above": self.above.parameters(),
             "dz": self.dz,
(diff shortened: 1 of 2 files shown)


bmaranville:
Tests are failing because classes that have `bumps.parameter.Constant` in their `parameters()` cause an infinite recursion when walking the parameter tree in `tag_all`. (`Constant.parameters()` returns `[self]`, so you can see where the recursion comes in...)

`Parameter.parameters()` also returns `[self,...]` but it doesn't cause infinite recursion because the tree-walker in `tag_all` finds that it has a `add_tag` method and that gets called instead of recursing through its `parameters()`.

Some solutions include

- skipping `Constant` when it is found in the `bumps.parameter.tag_all` walker
- returning `[]` from `Constant.parameters()`, which also seems reasonable
- adding tagging methods to `Constant`, which seems less reasonable

refl1d/deprecated/freeform.py
```diff
@@ -179,7 +179,7 @@ def _set_thickness(self, v):
 
     def parameters(self):
         return {
-            "interface": self.interface,
+            "interface": self.interface.parameters(),
```
pkienzle: Please remove this change.

If `"interface": self.interface` is an error, then this same error is pervasive.

For example, `dz` and `dp` are lists of parameters also defined by `Par.default`, so we would need `[p.parameters() for p in self.dz]`, etc.

bmaranville:
This edit was a mistake... since freeform isn't even used in our tests it was an edit in the wrong place.

pkienzle:
I don't think we need both instrument and nuisance tags covering exactly the same parameters.

@acaruana2009 can we drop the "nuisance" tag?

pkienzle:
Tests are failing because this PR needs https://github.com/bumps/bumps/pull/462

acaruana2009:
Hi @pkienzle, 

Firstly, I am happy to drop the nuisance tags. 

Just realised this is not the PR I was working on...

acaruana2009 (approved):

acaruana2009 (approved):

pkienzle:
I'm not clever enough to deal with tightly coupled bumps-refl1d tests, hence the cruft to allow a bit of version skew.

